PR #22, opened 2018-11-06
Merge request 

dede95:
Updated tier_1.json and added schema_transformation files.


nesta/production/schemas/tier_1/schema_transformations/eu_github.json
```diff
@@ -0,0 +1,19 @@
+[
```
jaklinger: This file looks like a duplicate of github_projects_and_users - is that true?

dede95: They are almost the same but the field names of the original datasets are different.

dede95: I named the .json files in a similar fashion to the datasets given to me.

PR #36, opened 2018-11-30
Juan files

Juan-Mateos:
Schema for kickstarter.

Some observations:

I have created a schema for the raw data before any processing.

-Non-standardised dates (from epoch) and locations
-Includes a few dicts (I created a dict firstName to deal with this)
-Includes a few empty / garbage fields (I created an empty firstName)
-What's the difference between date and datetime in the firstNames?

Thanks!

J


jaklinger:
- We will standardise any dates and locations before applying the schema mapping.
- Dict objects will have to be flattened out into entities, the only non-flat entity that we'll have in the schema will be a list
- garbage fields can be omitted
- The 'date' first_name is going to be deprecated in favour of 'datetime'

jaklinger:
Notes on 'kickstarter.json'

'*' : I've changed this
'#?': These may either need to be dropped or flattened (e.g. if a dict)
'##': These can probably be dropped altogether, as will never be used

Note: I'm going to close the pull request since you need to do 'git pull' before creating your branch 'juan_files' (see that you are committing 87 files. whereas you have only really changed 2 files). I'd git clone the repo fresh, and then checkout `dev` then checkout a new branch, copy this file over and then do the pull request.

PR #41, opened 2019-01-03
Crunchbase collection into mysql

russwinch:
Extraction of most of the data from the crunchbase tar file.

The Luigi pipeline consists of the following:
- extraction and processing of the organizations table. This was broken out due to the specific complexity of combining the parents data and long descriptions

- extraction and processing of all other tables (non-org), including conversion of any bools, creation of the location composite key where there is location data in the table

- geocoding, requiring a new class based on AutoBatchTask which manages a table in sql `geographic data`, adding new locations and geocoding them via batch. This is applied to the 5 crunchbase tables with cities and countries; 1 task per table. Some functionality from NIH was moved to the packages to enable this.

Crunchbase data schema here: https://docs.google.com/presentation/d/1e1aWpQ7M3klHc4iJXJhgpCD77uWFKaFTjVkyvdiNEXw/edit?usp=sharing

**Pytest generates multiple depreciation warnings when instantiating the luigi task, this seems to be based on the luigi codebase rather than ours. Hopefully this will still pass Travis**

*all data is extracted from the tar file, the geocoding task is still running*


nesta/packages/crunchbase/crunchbase_collect.py
```diff
@@ -0,0 +1,333 @@
...
+    Returns:
+        :code:`tarfile.Tarfile`: opened tar archive
+    """
+    crunchbase_config = misctools.get_config('crunchbase.config', 'crunchbase')
+    user_key = crunchbase_config['user_key']
+    url = 'https://api.crunchbase.com/v3.1/csv_export/csv_export.tar.gz?user_key='
+    with NamedTemporaryFile() as tmp_file:
+        r = requests.get(''.join([url, user_key]))
+        tmp_file.write(r.content)
+        try:
```
jaklinger: From the point of view of memory management, I think I would change this to:

       with NamedTemporaryFile() as tmp_file:
            r = requests.get(''.join([url, user_key]))
            tmp_file.write(r.content)
            tmp_tar = tarfile.open(tmp_file.name)

       try:
           yield tmp_tar
       finally:
           tmp_tar.close()

to allow `tmp_file` to go out of scope (though how useful this is depends how large the `tmp_file` is and for how long `tmp_tar` will be in scope)

nesta/packages/crunchbase/crunchbase_collect.py
```diff
@@ -0,0 +1,333 @@
...
+            tmp_tar = tarfile.open(tmp_file.name)
+            yield tmp_tar
+        finally:
+            tmp_tar.close()
+
+
+def get_csv_list():
```
jaklinger: Depends how much you want to... but it might be a good idea to refactor some of the guts of this function this into a shared package and call it `list_tarfile_contents(tarfile, regex_pattern=r'^(.*)$')`. Not sure how much it would get used, conceivably it could though.

nesta/packages/crunchbase/crunchbase_collect.py
```diff
@@ -0,0 +1,333 @@
...
+    with crunchbase_tar() as tar:
+        names = tar.getnames()
+    for name in names:
+        tablename = csv_pattern.match(name)
+        if tablename is not None:
+            csvs.append(tablename.group(1))
+    return csvs
+
+
+def get_files_from_tar(files, test=False):
```
jaklinger: Should `nrows` be an argument rather than `test`?

nesta/packages/crunchbase/crunchbase_collect.py
russwinch: temp file is huge >600MB! 

Good point, I will get this changed.

nesta/packages/crunchbase/crunchbase_collect.py
russwinch: @jaklinger with test as the kwarg I can just supply the luigi task's `self.test` variable. If I change it to nrows I'll need to perform `nrows = 1000 if self.test else None` each time before calling the function.

nesta/production/orms/crunchbase_orm.py
```diff
@@ -0,0 +1,249 @@
+'''
+Crunchbase
+======
```
jaklinger: sphynx will complain unless the `====` is at least as long as `Crunchbase`

nesta/production/luigihacks/tests/test_batchgeocode.py
```diff
@@ -0,0 +1,181 @@
...
+    geo_batch_task._put_batch = mock.Mock()
+    geo_batch_task.batch_size = 1000
+    geo_batch_task.test = False
+
+    uncoded_locations = create_uncoded_locations(2500)
+    batches = []
+    for batch in geo_batch_task._create_batches(uncoded_locations):
+        batches.append(batch)
+
+    assert len(batches) == 3
```
jaklinger: Add a comment here (and same in the test below this one too) explaining why `n == 3`

nesta/packages/crunchbase/crunchbase_collect.py
```diff
@@ -0,0 +1,333 @@
...
+        value (str): t or f
+
+    Returns:
+        (bool): boolean representation of the field, or None on failure
+    """
+    lookup = {'t': True, 'f': False}
+    try:
+        return lookup[value]
+    except KeyError:
+        logging.info(f"Not able to convert to a boolean: {value}")
```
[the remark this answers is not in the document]
russwinch: I needed to catch the error so I could just return None and the unconvertable data will just end up as a NULL in the database

nesta/packages/crunchbase/crunchbase_collect.py
```diff
@@ -0,0 +1,333 @@
...
+
+
+def _generate_composite_key(**kwargs):
```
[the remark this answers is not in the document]
russwinch: generate_composite_key is imported, but I felt this behaviour was specific to this package and not necessarily the way it would always want to be used

nesta/packages/crunchbase/crunchbase_collect.py
```diff
@@ -0,0 +1,333 @@
...
+        pks (list): names of primary key columns in the dataframe
+
+    Returns:
+        (:obj:`list` of :obj:`dict`): processed table with existing records removed
+    """
+    # fix uuid column names
+    df = rename_uuid_columns(df)
+
+    # drop any rows already existing in the database
+    total_records = len(df)
```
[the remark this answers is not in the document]
russwinch: done! how on earth did you notice that!?

nesta/production/luigihacks/tests/test_batchgeocode.py
russwinch: good idea, the second one had me confused when I read it back!

nesta/packages/crunchbase/crunchbase_collect.py
russwinch: need to add a test case to mock up a tar file

russwinch: Done!

nesta/packages/crunchbase/crunchbase_collect.py
russwinch: I am going to leave this for now

nesta/packages/crunchbase/crunchbase_collect.py
russwinch: test mode removed from package

nesta/packages/crunchbase/crunchbase_collect.py
```diff
@@ -0,0 +1,333 @@
...
+    Returns:
+        (str): composite key if able to generate, otherwise None
+    """
+    try:
+        return generate_composite_key(**kwargs)
+    except ValueError:
+        return None
+
+
+def total_records(data_dict, append_to=None):
```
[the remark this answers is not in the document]
russwinch: this code can probably be deleted as it's not really serving much function - too much data is returned and it's hard to debug.

PR #42, opened 2019-01-09
Crunchbase collect

russwinch:
I realised last night I had missed a crucial step in the mocking of the context manager...I hadn't closed the file when it __exit__'d so the test was invalid. It's a bit hacky as the side_effect tries to send too many args to the close function giving an error so I just captured them with *args in a lambda.

Test still passes but it's actually doing what I intended now!


russwinch:
removing this and re-doing from a new branch.

PR #46, opened 2019-01-15
Gtr geocode

russwinch:
*Geocoding currently running*

This populates the `gtr_organisations_locations` table with all organisation ids and *if possible* iso coding and geocoding.

The majority of organisations are in the UK but as the address data is from a multitude of sources it is inconsistent. As a result I am geocoding by postcode in the UK for this piece with the view to revisit it sometime in the near future and capture the rest.

This PR includes:
- Moving the split batch functionality to misc_utils
- Improving the geocoding wrapper so all allowed query options can be sent, not just city+country
- ORM for the `gtr_organisations_locations` table
- Luigi task


russwinch:
@jaklinger Geocoding now finished 👍 

Summary of 38k organisations:
20k with geocoding
23k with country codes
------------------------
~15k with neither...to hopefully be picked up in the *hard* geocoding task

PR #56, opened 2019-02-26
Add parameter to batch geocoder to choose query method

russwinch:
There is now a flag enabling the query method to be chosen. I went with 0, 1, 2 as the options over string representations.

Tests added etc.

Geocoding using this method have been re-run and the results look promising: correct for Nottingham which was incorrect before and there are only ~500 which failed to geocode.

I foolishly truncated the table before re-running it rather than making a copy *facepalm* but it only took 10 hours to run so we can go back in less than a day :oO Kostas is going to have a look at the data when he can as he noticed the issue during his analysis.

### DIFF
--- a/nesta/packages/geo_utils/geocode.py
+++ b/nesta/packages/geo_utils/geocode.py
@@ -101,7 +101,8 @@ def geocode_dataframe(df):
 
 
 def geocode_batch_dataframe(df, city='city', country='country',
-                            latitude='latitude', longitude='longitude'):
+                            latitude='latitude', longitude='longitude',
+                            query_method=2):
     """Geocodes a dataframe, first by supplying the city and country to the api, if this
     fails a second attempt is made supplying the combination using the q= method.
     The supplied dataframe df is returned with additional columns appended, containing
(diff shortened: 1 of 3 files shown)


nesta/packages/geo_utils/geocode.py
```diff
@@ -113,15 +114,23 @@ def geocode_batch_dataframe(df, city='city', country='country',
         country (str): name of the input column containing the country
         latitude (str): name of the output column containing the latitude
         longitude (str): name of the output column containing the longitude
+        query_method (int): query methods to attempt: 0: city, country only
```
jaklinger: Apologies for the delay here in reviewing. I think the 0,1,2 method might leave us vulnerable to difficult debugging in the future. I'd favour using string labels unless performance is an issue - whereas the bottleneck in this case is really the geocode API - so ['city_country_only', 'query_only', 'both'] might be more useful. I don't think that would take long to implement right?

russwinch: This makes sense, it was already annoying to remember what each code meant while I was just changing them over

PR #68, opened 2019-04-16
Arxiv iterative

russwinch:
Here is the arxiv iterative collection pipeline, with MAG query for fields of service.

This became more complicated than I expected, some of which was self inflicted but I think it makes sense. Of particular note: I felt the MAG model would likely be used for a variety of datasets and each would implement it's own link table to the FieldsOfStudy table; as a result the FieldsOfStudy table would need to be in it's own .py file and sharing of the Base then became a problem. I've implemented a simple metadata merger so Base.metadata.create_all() will now work as well as the relationships which I finally got working.

Althrough great for querying the relationships are very slow to write, so these have functions to write using the SQLAlchemy core orm, which is so much faster.

Potentially controversial are the two OOP implementations in the arxiv packages:
_BatchedTitles_: as you'll see there is a nest of generators stacked here but I needed to yield titles while still providing access to a lookup table which was changing dynamically for each batch requested from the database. I initially solved it by passing another variable into the function, but it felt nasty so I switched to a class.
_BatchWriter_: this neatened up the duplicate code I was suffering from the batch writing in through the pipeline. May be terrible. Please advise.

**The non iterative collect is effectively superseded by this version. Although it used the batch process, this approach is so much more efficient at query and write that I would use it over the last one to re-collect all data, if necessary.**

_The pipeline has been run and the data is up to date. It looks like there are going to be some merge conflicts, I'll add additional commits to resolve these_


nesta/packages/arxiv/collect_arxiv.py
```diff
@@ -201,15 +212,221 @@ def arxiv_batch(token, cursor):
...
+        self.limit = limit
+        self.function = function
+        self.function_args = function_args
+        self.function_kwargs = function_kwargs
+
+    def append(self, item):
+        """Appends to the list and then checks current size against the set limit."""
+        super().append(item)
+        if len(self) >= self.limit:
+            self.write(self.copy())
```
jaklinger: what's the purpose of `copy` here?

nesta/packages/arxiv/collect_arxiv.py
```diff
@@ -201,15 +212,221 @@ def arxiv_batch(token, cursor):
...
+            for article in (self.session.query(Article)
+                                        .filter(Article.id.in_(batch_of_ids))
+                                        .all()):
+                self.title_articles_lookup[prepare_title(article.title)].append(article.id)
+
+            for title in self.title_articles_lookup:
+                yield title
+
+
+class BatchWriter(list):
```
jaklinger: nice - this is probably better off in a `misc_utils` or similar. what do you think?

nesta/packages/mag/query_mag.py
```diff
@@ -0,0 +1,337 @@
...
+        fields: (:obj:`list` of `str`): codes of fields to return, as per mag documentation
+        query_count: (int): number of items to return
+        offset (int): offset in the results if paging through them
+
+    Returns:
+        (dict): json response from the api containing 'expr' (the original expression)
+                and 'entities' (the results) keys.
+                If there are no results 'entities' is an empty list.
+    """
+    endpoint = "https://api.labs.cognitive.microsoft.com/academic/v1.0/evaluate"
```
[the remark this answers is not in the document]
russwinch: I have fixed this in the other branch as the file gets split in 2

nesta/packages/mag/query_mag.py
```diff
@@ -0,0 +1,337 @@
...
+            yield query_prefix_format.format(','.join(expr))
+            expr.clear()
+        expr.append(formatted_item)
+
+    # pick up any remainder below max_length
+    if len(expr) > 0:
+        yield query_prefix_format.format(','.join(expr))
+
+
+def query_mag_api(expr, fields, subscription_key, query_count=1000, offset=0):
```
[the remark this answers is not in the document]
russwinch: great! done

nesta/packages/arxiv/collect_arxiv.py
russwinch: the copy was there to make it testable 😞 if I don't copy it before the list gets cleared my test fails as the values are no longer in the call to my mock. I would love to find a better way to solve this!

russwinch: after changing to your version of the code above it is passing the test fine now 👍

nesta/packages/mag/query_mag.py
```diff
@@ -0,0 +1,337 @@
...
+        # print(expr)
+        data = query_mag_api(expr, paper_fields, subscription_key)
+        # print(json.dumps(data['entities'][0], indent=4))
+        print(f"query length: {len(expr)}")
+        print(f"titles in query: {len(expr.split(','))}")
+        print(f"entities returned from api: {len(data['entities'])}")
+
+        break
+
+    # for row in data['entities']:
```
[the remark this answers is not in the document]
russwinch: I have dramatically tidied up in this area, it was bad. I moved the data science code to a gist.

nesta/packages/arxiv/collect_arxiv.py
```diff
@@ -201,15 +212,221 @@ def arxiv_batch(token, cursor):
...
+            self.clear()
+
+    def extend(self, items):
+        """Adds multiple items to the list then writes until it is below limit."""
+        super().extend(items)
+        while len(self) >= self.limit:
+            self.write(self[0:self.limit])
+            del self[0:self.limit]
+
+    def write(self, items=None):
```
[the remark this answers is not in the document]
russwinch: awesome! passing self[0:n] to the function solved the bug I was having with the test, removing the need for that hacky `.copy()` below

nesta/packages/arxiv/collect_arxiv.py
russwinch: moved to `misc_utils.batches`

russwinch:
I think this should be ready to go, I'm going to try and merge it into the branch I have been working on as well, to deal with any conflicts before I forget which version is correct.

PR #74, opened 2019-05-03
Change mapping and aliases for NIH and CB

### LINKED ISSUE #71: Change mapping and aliases for NIH and CB
Some fields would be better off as `text` to be able to search in them:
- places in general
- websites (excluding social media as these usually have IDs or nicknames in URL, TBD)
- categories: a company might be categorised as "medical devices" and ES (or Clio?) might rank it higher when we are searching for "medicine".

We need a `state_id` for NIH too.

`countries_codes` should be `countries_ids` to be consistent with a single `countries_id` (and it's shorter).

mindrones:
- [x] Rename 'countries_codes', add 'state_id'
- [x] change some fields type from keyworkd -> text
- [x] sort orms configs alphabetically helps reading
- [x] check the [array fields](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/array.html)
- [ ] squash

Closes #71

### DIFF
--- a/nesta/production/orms/nih_es_config.json
+++ b/nesta/production/orms/nih_es_config.json
@@ -17,7 +17,8 @@
                 }
             }
         }
-    },"mappings": {
+    },
+    "mappings": {
         "_doc": {
             "dynamic": "strict",
             "properties": {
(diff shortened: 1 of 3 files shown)


jaklinger:
sitting on this review until after the new data processing has been checked

mindrones:
See https://github.com/nestauk/nesta/issues/81#issuecomment-498218507

nesta/production/orms/nih_es_config.json
```diff
@@ -86,31 +58,80 @@
...
                             "type": "keyword"
                         }
-                    },
-                    "analyzer": "mesh_terms_analyzer"
+                    }
                 },
-                "booleanFlag_duplicate_abstract": {
-                    "type": "boolean",
-                    "null_value": false
+                "placeName_continent_organisation": {
```
jaklinger: continents are single words, so I guess this is redundant

mindrones: The reason I added `text`here is that if I got it right with `keyword` looking for "asia" won't find ["Asia"](https://nesta-open-data.s3.eu-west-2.amazonaws.com/rwjf-viz/iso_a2_to_name_by_type.json), correct?

jaklinger: good point!

jaklinger: oh wait, no your logic is the wrong way around: `keyword` is case insensitive so "Asia" and "asia" will both return "Asia"

mindrones: Ah I see, good to know! Then I'd say `text` can go

jaklinger:
ok to close this?

mindrones:
yeps 👍

PR #83, opened 2019-05-24
Grid

russwinch:
A task is added to the arXiv pipeline to link GRID to arXiv articles through MAG authors and affiliations.

Exact matches are sought using GRID id, institute name and then falling back to fuzzy matching.

An additional query has been added into the MAG sparql task to collect author information, as this also contains GRID id.

ComboFuzzer is added from the arxiv_ai research and it has a new method added to perform the fuzzy matching and keep track of previous history.


nesta/packages/arxiv/collect_arxiv.py
```diff
@@ -350,13 +350,20 @@ def add_new_articles(article_batch, session):
...
     Args:
         article_batch (:obj:`list` of `dict`): articles to add to database
         session (:obj:`sqlalchemy.orm.session`): active session to use
     """
     logging.info(f"Updating a batch of {len(article_batch)} existing articles")
 
+    # look for any institutes in provided article_batch
+    for article in article_batch:
+        if article.get('institutes'):
+            raise TypeError("Institute links cannot be written using this method. Use add_article_institutes instead")
```
russwinch: On second thoughts this should possibly be `NotImplementedError`?

nesta/packages/mag/query_mag_sparql.py
```diff
@@ -224,6 +259,61 @@ def update_field_of_study_ids_sparql(session, fos_ids):
...
+    '''
+    concat_format = "<http://ma-graph.org/entity/{}>"
+
+    logging.debug(f"Querying MAG for authors: {ids}")
+    for count, row in enumerate(_batch_query_sparql(query,
+                                                    concat_format=concat_format,
+                                                    filter_on='author',
+                                                    ids=ids), start=1):
+        # rename and reformat fields out of entity urls
+        row['author_id'] = extract_entity_id(row.pop('author'))
```
jaklinger: might be better to have this as an old-new field name mapping `dict`, which is an arg to `query_authors`.

nesta/packages/grid/grid.py
```diff
@@ -0,0 +1,182 @@
...
+    """Reads the grid.csv and addresses.csv flatfiles, loads it into a dataframe and
+    combines them, dropping duplicate and redundant data.
+
+    Args:
+        filepath (str): location of the folder containg the extracted files
+
+    Returns
+        (:obj:`pandas.DataFrame`): all data from the file
+    """
+    df = pd.read_csv(f"{filepath}/grid.csv", low_memory=False)
```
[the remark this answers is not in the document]
russwinch: as discussed, GRID is not currently part of any pipeline, just the creation of the links between arxiv and GRID. Ticket created to resolve this at some point in the future and docstring updated to clarify.

nesta/packages/mag/query_mag_sparql.py
russwinch: I've added the dict which made this look much neater, but I decided to keep it within the function for now as it's only relevant to the query written within that function. To be reviewed in future if there are more queries added.

nesta/packages/mag/query_mag_sparql.py
```diff
@@ -94,70 +102,71 @@ def query_mag_sparql_by_doi(articles):
...
+            entities = ','.join(concat_format.format(i) for i in batch_of_ids)
+            entity_filter = f"FILTER (?{filter_on} IN ({entities}))"
+    elif any([concat_format, filter_on, ids]):
+        raise ValueError("concat_format, filter_on and ids must all be supplied together or not at all")
+    else:
+        # retrieve all fields of study
+        entity_filter = ''
+
+    for batch in sparql_query(MAG_ENDPOINT, query.format(entity_filter)):
+        yield from batch
```
russwinch: I changed this from a nested for loop. been wanting to try yield from out!

PR #86, opened 2019-05-28
Data Quality- Missing Values merge

dede95:


nesta/packages/quality_check/quality_calculations.py
```diff
@@ -0,0 +1,68 @@
+import pandas as pd
+import itertools
+
+def missing_values(data_frame):
```
georgerichardson: This is good, but everything needs a docstring.

nesta/packages/quality_check/quality_calculations.py
```diff
@@ -0,0 +1,68 @@
...
+
+def missing_values(data_frame):
+
+    """
+
+    """
+    return data_frame.isnull().sum().sort_values(ascending=False)
+
+def missing_value_column_count(data_frame):
+
```
georgerichardson: Might be a bit more readable to use numpy's `histogram` function

```suggestion
data_frame_missing = data_frame.isnull().sum() / data_frame.shape[0]
out = np.histogram(data_frame_missing, bins=range(0, 101, 10))
return out
```

nesta/packages/quality_check/quality_calculations.py
```diff
@@ -0,0 +1,68 @@
...
+    bins_list = [i for i in range(100+1) if i%10 == 0]
+    out = pd.cut(data_frame.isnull().mean().apply(lambda x: round(100*x,5)), bins=bins_list, include_lowest=True)
+
+    return out.value_counts(sort=False)
+
+def missing_value_count_pair_both(data_frame):
+#include here or in script itself
+    pair_list = []
+    for pair in itertools.combinations_with_replacement(data_frame.columns,2):
```
georgerichardson: This first loop is redundant.

nesta/packages/quality_check/quality_calculations.py
```diff
@@ -0,0 +1,68 @@
...
+        pair_list.append(pair)
+        #1min 30secs
+    both_null_dict = {}
+    for pair in pair_list:
+        bool_1 = pd.isnull(data_frame[pair[0]])
+        bool_2 = pd.isnull(data_frame[pair[1]])
+
+        and_counts = (bool_1 & bool_2)
+        if True in and_counts.value_counts().index:
```
georgerichardson: This line probably slows this function down quite a bit
```suggestion
        if True in and_counts.values:
```

To be honest, you could probably just skip this if statement and just calculate the sum for all pairs. It should be very fast to calculate on a boolean array even for millions of rows.

nesta/packages/quality_check/quality_calculations.py
```diff
@@ -0,0 +1,68 @@
...
+            both_null_dict[(pair[0],pair[1])] = and_counts.sum()
```
georgerichardson: This is the same as:
```suggestion
            both_null_dict[pair] = and_counts.sum()
```

nesta/packages/quality_check/quality_calculations.py
```diff
@@ -0,0 +1,68 @@
...
+    pair_list = []
+    for pair in itertools.combinations_with_replacement(data_frame.columns,2):
+        pair_list.append(pair)
+
+    either_null_dict = {}
+    for pair in pair_list:
+        bool_1 = pd.isnull(data_frame[pair[0]])
+        bool_2 = pd.isnull(data_frame[pair[1]])
+
+        either_counts = ((bool_1 ==True) | (bool_2==True))
```
georgerichardson: Don't need to evaluate whether a boolean array is True. Also might as well just calculate the sum once.

```suggestion
        either_counts = (bool_1 | bool_2).sum()
```

nesta/packages/quality_check/quality_calculations.py
```diff
@@ -0,0 +1,68 @@
...
+    pair_list = []
+    for pair in itertools.combinations_with_replacement(data_frame.columns,2):
+        pair_list.append(pair)
+        #1min 30secs
+    both_null_dict = {}
+    for pair in pair_list:
+        bool_1 = pd.isnull(data_frame[pair[0]])
+        bool_2 = pd.isnull(data_frame[pair[1]])
+
+        and_counts = (bool_1 & bool_2)
```
georgerichardson: Would just calculate the sum here once.
```suggestion
        and_counts = (bool_1 & bool_2).sum()
```

nesta/packages/quality_check/quality_calculations.py
```diff
@@ -0,0 +1,68 @@
...
+            both_null_dict[(pair[0],pair[1])] = and_counts.sum()
+            both_null_dict[(pair[1],pair[0])] = and_counts.sum()
+
+        else:
+            both_null_dict[(pair[0],pair[1])] = 0
+            both_null_dict[(pair[1],pair[0])] = 0
+
+    return both_null_dict
+
+def missing_value_count_pair_either(data_frame):
```
georgerichardson: General comments from the previous function apply here too.

nesta/packages/quality_check/quality_calculations.py
```diff
@@ -0,0 +1,68 @@
...
+            either_null_dict[(pair[0],pair[1])] = either_counts.sum()
+            either_null_dict[(pair[1],pair[0])] = either_counts.sum()
+
+        else:
+            either_null_dict[(pair[0],pair[1])] = 0
+            either_null_dict[(pair[1],pair[0])] = 0
+
+    return either_null_dict
+
+def melt_dict_to_df(calc_dict):
```
georgerichardson: Maybe make this function 'private' (put a '_' at the start of its name) and use it inside the missing pairs functions so that they return DataFrames instead of dicts.

nesta/packages/quality_check/quality_calculations.py
georgerichardson: This is good, but everything needs a docstring.

nesta/packages/quality_check/quality_calculations.py
```diff
@@ -0,0 +1,68 @@
...
+
+    """
+    return data_frame.isnull().sum().sort_values(ascending=False)
+
+def missing_value_column_count(data_frame):
+
+    bins_list = [i for i in range(100+1) if i%10 == 0]
+    out = pd.cut(data_frame.isnull().mean().apply(lambda x: round(100*x,5)), bins=bins_list, include_lowest=True)
+
+    return out.value_counts(sort=False)
```
georgerichardson: Might be a bit more readable to use numpy's `histogram` function

```suggestion
data_frame_missing = data_frame.isnull().sum() / data_frame.shape[0]
out = np.histogram(data_frame_missing, bins=range(0, 101, 10))
return out
```

nesta/packages/quality_check/quality_calculations.py
georgerichardson: This first loop and the list of pairs is redundant. Just add the pairs as keys to the null_dict as they are generated.

nesta/packages/quality_check/quality_calculations.py
georgerichardson: Would just calculate the sum here once.
```suggestion
        and_counts = (bool_1 & bool_2).sum()
```

nesta/packages/quality_check/quality_calculations.py
georgerichardson: This line probably slows this function down quite a bit
```suggestion
        if True in and_counts.values:
```

To be honest, you could probably just skip this if statement and just calculate the sum for all pairs. It should be very fast to calculate on a boolean array even for millions of rows.

nesta/packages/quality_check/quality_calculations.py
georgerichardson: This is the same as:
```suggestion
            both_null_dict[pair] = and_counts.sum()
```

nesta/packages/quality_check/quality_calculations.py
georgerichardson: General comments from the previous function apply here too.

georgerichardson: Actually, now that I think about it. Why not have a single function for these? You could just have an extra argument that takes 'and' or 'or' and just chooses between `(bool_1 & bool_2).sum()` and `(bool_1 | bool_2).sum()` depending on the user's choice.

nesta/packages/quality_check/quality_calculations.py
georgerichardson: Don't need to evaluate whether a boolean array is True. Also might as well just calculate the sum once.

```suggestion
        either_counts = (bool_1 | bool_2).sum()
```

nesta/packages/quality_check/quality_calculations.py
```diff
@@ -0,0 +1,68 @@
...
+
+def missing_value_count_pair_both(data_frame):
+#include here or in script itself
+    pair_list = []
+    for pair in itertools.combinations_with_replacement(data_frame.columns,2):
+        pair_list.append(pair)
+        #1min 30secs
```
georgerichardson:
```suggestion
```

nesta/packages/quality_check/quality_calculations.py
```diff
@@ -0,0 +1,68 @@
...
+
+def missing_value_column_count(data_frame):
+
+    bins_list = [i for i in range(100+1) if i%10 == 0]
+    out = pd.cut(data_frame.isnull().mean().apply(lambda x: round(100*x,5)), bins=bins_list, include_lowest=True)
+
+    return out.value_counts(sort=False)
+
+def missing_value_count_pair_both(data_frame):
+#include here or in script itself
```
georgerichardson:
```suggestion
```

nesta/packages/quality_check/quality_function.py
```diff
@@ -0,0 +1,83 @@
+from quality_calculations import *
```
georgerichardson: Very minor point, but there is a convention for import orders which is typically
```
import <built in python modules>

import <external packages>

import <internal code>
```

jaklinger:
No activity for a few months so closing.

Issue #96, opened 2019-06-04
Meetup data processing for rwjf


jaklinger:
Relating to branch https://github.com/nestauk/nesta/tree/96-rwjf_mup

- [x] Devise method for automatic identification of health related meetup groups
- [x] Add local task for extracting topics relating to health
- [x] Add geocoding into pipeline
- [x] Pipe relevant meetup data to ES
- [x] Add `terms` alias for `terms_topics_group`
- [x] Add auto-translation
- [x] Sanitise html

Also related to #102

jaklinger:
Index `meetup_v0` is now available https://search-health-scanner-5cs7g52446h7qscocqmiky5dn4.eu-west-2.es.amazonaws.com/meetup_v0/_search

mindrones:
Closed by #122, moved the remaining points to #126

PR #99, opened 2019-06-06
96 rwjf mup

jaklinger:
Task for selecting, enriching, cleaning and piping meetup data to ES.

See discussion here:
#96


nesta/packages/meetup/meetup_utils.py
```diff
@@ -59,13 +64,56 @@ def flatten_data(list_json_data, keys, **kwargs):
...
+        engine: A SQL alchemy connectable.
+        perc (int): A percentile to evaluate.
+    Returns:
+        members (float): The number of members corresponding to this percentile.
+    """
+    with db_session(engine) as session:
+        rows = (session
+                .query(Group.members)
+                .all())
+        rows = [r[0] for r in rows]
```
russwinch: If this query is very slow already then forget it, otherwise id probably get rid of that index:
```
        rows = (session
                .query(Group)
                .all())
        rows = [r.members for r in rows]
```
or maybe even just:
`rows = [g.members for g in session.query(Group).all()]`

It seems to be (slightly) slower than with the index

nesta/packages/meetup/meetup_utils.py
```diff
@@ -59,13 +64,56 @@ def flatten_data(list_json_data, keys, **kwargs):
...
+                             for it to be considered.
+        perc (int): A percentile to evaluate the most frequent topics.
+    Returns:
+        topics (set): The set of most frequent topics.
+    """
+    with db_session(engine) as session:
+        rows = (session
+                .query(Group.topics)
+                .filter(Group.members >= members_limit)
+                .filter(Group.category_shortname.in_(core_categories))
```
russwinch: I didn't realise you could combine multiple `.filter` like this for AND, I have been using ugly multi-line bitwise operations 👍

nesta/packages/meetup/meetup_utils.py
```diff
@@ -59,13 +64,56 @@ def flatten_data(list_json_data, keys, **kwargs):
...
+                .all())
+        rows = [r[0] for r in rows]
```
russwinch: remove index if it's not implemented for slight performance gains

nesta/production/luigihacks/estask.py
```diff
@@ -0,0 +1,132 @@
...
+from nesta.packages.misc_utils.batches import split_batches, put_s3_batch
+from nesta.production.luigihacks import autobatch
+from nesta.production.luigihacks.misctools import get_config
+from nesta.production.luigihacks.mysqldb import MySqlTarget
+from nesta.production.orms.orm_utils import get_mysql_engine
+from nesta.production.orms.orm_utils import setup_es
+from nesta.production.orms.orm_utils import db_session
+
+
+class ElasticsearchTask(autobatch.AutoBatchTask):
```
russwinch: looking forward to using this 👍

nesta/production/orms/meetup_es_config.json
```diff
@@ -0,0 +1,132 @@
...
+                },
+                "placeName_state_group": {
+                    "fields": {
+                        "keyword": {
+                            "type": "keyword"
+                        }
+                    },
+                    "type": "text"
+                },
+                "terms_memberOrigin_group": {
```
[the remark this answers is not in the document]
jaklinger: it's a field that doesn't exist yet. I'm going to add a field describing the country origins of the members, above a threshold. For example, if 90% of the members are from France, 8% are from Germany and 2% from elsewhere, a threshold of 5% would mean that `terms_memberOrigin_group` would be equal to `['FR', 'DE']` which would allow you to map the 'internationalness' of the group. This is briefly indicated in the task list of issue #96 

I'm going to prefix the field name with `_terms` to clarify that this is in development.

nesta/production/schemas/tier_1/aliases/health_scanner.json
```diff
@@ -78,7 +86,8 @@
     },
     "state": {
         "nih": "placeName_state_organisation",
-        "crunchbase": "placeName_state_organisation"
+        "crunchbase": "placeName_state_organisation",
+        "meetup": "placeName_state_group"
     },
     "state_id": {
         "nih": "id_state_organisation",
```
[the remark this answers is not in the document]
jaklinger: meetup doesn't have a `state` field unfortunately. This could be added in in the future, but will require some new dev for that. I've realised that `meetup_es_config.json` indicates that there is, so I've prefixed these with an underscore now.

nesta/packages/meetup/meetup_utils.py
jaklinger: It's not just slower, the memory requirement will also be fairly large since it will fetch the entire row (such as the description which is a `TEXT` field). Although your suggestion looks much prettier, I'd prefer to stick to the performant once since I've had `OutOfMemory` exceptions with meetup data in the past (especially if running on the wee batch machines!)

nesta/packages/meetup/meetup_utils.py
jaklinger: functional programming ftw

nesta/production/orms/meetup_es_config.json
```diff
@@ -0,0 +1,132 @@
...
+                }
+            }
+        }
+    },
+    "settings": {
+        "analysis": {
+            "analyzer": {
+                "mesh_terms_analyzer": {
+                    "filter": [
+                        "standard",
```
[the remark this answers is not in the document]
jaklinger: I can't think of a case where order matters, so I will add sorting by default to `terms` being processed by `ElasticsearchPlus`. I'll add that to this PR, but you won't see changes in the data until the next reprocessing.

nesta/packages/meetup/meetup_utils.py
russwinch: indeed it does. I didn't think it executed the query until you actually interrogated a field within the object but it's actually `select *`'ing the whole row.... I have some serious optimisation to go back and do in my tasks...

jaklinger: the `all()` does the execution. Doing a `fetchXXX()` command would reduce the memory overhead but obviously requires more requests, so will be considerably slower. I think the syntax you choose should be use case dependent

russwinch: you should still be able to do `.members` on it in place of [0] as it is a NamedTuple

russwinch: not that there is much point here, but I have just discovered this, in case it is useful when querying multiple fields

jaklinger: ah - interesting, didn't know that. I'll put that in. Thanks!

nesta/production/orms/meetup_es_config.json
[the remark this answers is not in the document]
mindrones: wondering if these can be [ES objects](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/object.html), like `{FR: 0.9, DE: 0.08}`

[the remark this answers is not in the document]
jaklinger: certainly could be - perhaps both fields would be useful from a performance perspective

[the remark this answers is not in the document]
mindrones: like adding a field `object_memberOrigin_group` (object) on top of `terms_memberOrigin_group` (array) ?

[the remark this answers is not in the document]
jaklinger: yep - I've just added these to issue #96. Good to sign off?

nesta/production/orms/meetup_es_config.json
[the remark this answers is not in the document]
mindrones: ah, I meant sort arrays in these config files too

nesta/production/schemas/tier_1/aliases/health_scanner.json
```diff
@@ -1,15 +1,18 @@
 {
```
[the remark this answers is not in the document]
mindrones: the `url` alias is still missing

PR #106, opened 2019-06-10
Initial containerisation of Luigi

russwinch:
This includes the first version of Dockerised Luigi where container images are built from local files. 

The arXlive pipeline needed minor amendments to make it run (relative import for `requires` swapped to absolute path)

find_filepath_from_pathstub also needed a slight amendment as it was failing when the starting directory was one level down from HOME.


docker/Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+# Setup the virtualenv
+RUN python -m venv /venv
+# or "virtualenv /venv" for Python 2
+
+# Install Python deps
+# for dev copy locally, or specify a branch
+COPY requirements.txt ./
+RUN pip install --no-cache-dir -r requirements.txt
+
+# implement pull from github for production version (branch or tag can be supplied)
```
jaklinger: Are these comments needed for future dev, or safe to go?

russwinch:
I've tidied up the Dockerfile and have added python version as an optional argument during the build. Docs are updated to reflect this.

PR #107, opened 2019-06-10
Added url alias for mup

jaklinger:


nesta/production/schemas/tier_1/aliases/health_scanner.json
```diff
@@ -108,4 +108,8 @@
     "title": {
         "nih": "title_of_project"
     }
```
mindrones: missing a comma I think?

PR #113, opened 2019-06-25
Cordis fp7 h2020

jaklinger:
Written ORM for h2020 and data collection script (no tests yet, but useful for comparing methods to FP7).


nesta/packages/misc_utils/guess_sql_type.py
```diff
@@ -0,0 +1,27 @@
...
+    Returns:
+        sql_type (str): SQL type name
+    '''
+    _types = [type(row) for row in df_col
+             if not pd.isnull(row)]
+    # Assign type by str --> float --> int --> bool hierarchy
+    _type = (str if str in _types 
+             else (float if float in _types 
+                   else (int if int in _types
+                         else (bool if bool in _types else None))))
```
russwinch: would this be any better as:
```
types_hierarchy = [str, float, int, bool]
_type = None
for type in types_hierarchy:
    if type in _types:
        _type = type
        break
```
?
You could pass the hierarchy as an argument then

nesta/packages/misc_utils/guess_sql_type.py
```diff
@@ -0,0 +1,27 @@
...
+    if _type is str:
+        _len = max(len(str(row)) for row in df_col if not pd.isnull(row))
+        _type = 'TEXT' if _len > text_len else f'VARCHAR({_len})'
+    else:
+        _type = f'{lookup[_type]}'
```
russwinch: I can't see why this needs to be converted to a string, it's already a string?

nesta/production/orms/cordis_h2020_orm.py
```diff
@@ -0,0 +1,126 @@
...
+    post_code = Column(VARCHAR(20, collation='utf8_bin'))
+    organization_url = Column(TEXT(collation='utf8_bin'))
+    vat_number = Column(VARCHAR(20, collation='utf8_bin'))
+    contact_form = Column(TEXT(collation='utf8_bin'))
+    contact_type = Column(VARCHAR(14, collation='utf8_bin'))
+    contact_title = Column(VARCHAR(5, collation='utf8_bin'))
+    contact_first_names = Column(VARCHAR(12, collation='utf8_bin'))
+    contact_last_names = Column(VARCHAR(14, collation='utf8_bin'))
+    contact_telephone_number = Column(VARCHAR(17, collation='utf8_bin'))
+    contact_fax_number = Column(VARCHAR(17, collation='utf8_bin'))
```
russwinch: would it make more sense just to set the whole table collation? ` __table_args__ = {'mysql_collate': 'utf8_bin'}
` 

It will mean all the tables have to be collated otherwise the foreign key constraints can't be set up. 

I appreciate that it may take up more space, so maybe not worth it, but it would be neater.

nesta/production/orms/cordis_h2020_orm.py
```diff
@@ -0,0 +1,126 @@
...
+    rcn = Column(INTEGER, primary_key=True)
+    title = Column(TEXT(collation='utf8_bin'))
+    #project_id = Column(INTEGER)
+    #project_acronym = Column(TEXT(collation='utf8_bin'))
+    #programme = Column(TEXT(collation='utf8_bin'))
+    #topics = Column(VARCHAR(37))
+    authors = Column(TEXT(collation='utf8_bin'))
+    journal_title = Column(TEXT(collation='utf8_bin'))
+    journal_number = Column(TEXT(collation='utf8_bin'))
+    published_year = Column(INTEGER, primary_key=True)
```
russwinch: does this need to be a primary key if `last_update_date` is also a pk?

Issue #102, opened 2019-06-06
health scanner: second reprocessing


mindrones:
- [x] sort string arrays alphabeticallty: `terms`, `countries_ids`, `funders` (important to speed up building networks and helps readability if they are presented as lists): IMO this should be standard in ES+ for string arrays. (marked as done as, as discussed, this is done [here](https://github.com/nestauk/nesta/blob/dev/nesta/production/luigihacks/elasticsearchplus.py#L113))
- [x] rename meetups [`datetime_start_group`](https://github.com/nestauk/nesta/blob/dev/nesta/production/orms/meetup_es_config.json#L38) to `date_start_group` as it's `"yyyy-MM-dd"`
- [x] meetup has `id_iso{2|3}_group` as opposed to CB and NIH having `id_iso{2|3}_country`: we agreed to rename `id_iso{2|3}_country` on CB and NIH
- [x] rename meetup's `type_of_entity` from `"meetup group"` to `"meetup"`
- [x] indent here https://github.com/nestauk/nesta/blob/dev/nesta/production/schemas/tier_1/aliases/health_scanner.json#L112

Related to #81

mindrones:
Closed by #122

PR #122, opened 2019-07-17
Issues 96a & 102a (RWJF)

jaklinger:
Referring to issues #96 and #102 the following features have been implemented:

For meetup:
- Add terms alias for `terms_topics_group`
- Add auto-translation (including flag `booleanFlag_autotranslate_entity`)
- Sanitise html
- rename meetup's type_of_entity from "meetup group" to "meetup"
- indent here (health_scanner.json)
- rename meetups `datetime_start_group` to `date_start_group` as it's "yyyy-MM-dd"
- meetup has `id_iso{2|3}_group` as opposed to CB and NIH having `id_iso{2|3}_country`: we agreed to rename `id_iso{2|3}_country` on CB and NIH

New index `meetup_v2` is available. I can switch this to be aliased under `health_scanner` once this is signed off.

Interesting note: amount of data translated from another language gives 17% increase in meetup groups which can be searched (which should roughly all be non-english speaking countries or regions)


nesta/production/orms/meetup_es_config.json
```diff
@@ -29,13 +29,16 @@
                 "_terms_sdg_description": {
                     "type": "keyword"
                 },
+                "booleanFlag_autotranslate_entity": {
```
mindrones: minor, I'd rename to `booleanFlag_autotranslated_entity` or `bool_transl_entity` (the shorter the better [*])

nesta/production/orms/meetup_es_config.json
```diff
@@ -45,13 +48,13 @@
                 "id_country_group": {
                     "type": "keyword"
                 },
-                "id_iso2_group": {
```
mindrones: I think `id_iso2_group`, `id_iso3_group` and `id_isoNumeric_group` were supposed to be correct while for CB and NIH we were supposed to do:

- `id_iso2_country` -> `id_iso2_organisation`
- `id_iso3_country` -> `id_iso3_organisation`
- `id_isoNumeric_country` -> `id_isoNumeric_organisation`

See (*)

nesta/production/schemas/tier_1/aliases/health_scanner.json
```diff
@@ -39,7 +39,7 @@
     },
     "country_id": {
         "crunchbase": "id_iso2_country",
-        "meetup": "id_iso2_group",
```
mindrones: See (*)

nesta/production/schemas/tier_1/schema_transformations/meetup.json
```diff
@@ -18,15 +18,15 @@
     },
     {
         "tier_0": "country",
-        "tier_1": "id_iso2_group"
+        "tier_1": "id_iso2_country"
```
mindrones: See (*)

mindrones:
Actually, I realised there was a mistake in my review:

> `is_en` | `is_english` [*] -> `booleanFlag_autotranslated_entity`

should be:

`non_en` | `non_english` [*] -> `booleanFlag_autotranslated_entity`

or so.

jaklinger:
> Actually, I realised there was a mistake in my review:
> 
> > `is_en` | `is_english` [*] -> `booleanFlag_autotranslated_entity`
> 
> should be:
> 
> `non_en` | `non_english` [*] -> `booleanFlag_autotranslated_entity`
> 
> or so.

I'm not too fussed personally what the alias is, however I'm not sure that non-english is quite right. The boolean is to indicate that the text has been translated to english from another language. perhaps `is_translated` or just `translated`?

mindrones:
`translated` would be perfect

nesta/production/luigihacks/elasticsearchplus.py
```diff
@@ -19,15 +23,144 @@
...
+
+def strip_tags(html):
+    """Taken from https://stackoverflow.com/questions/753052/strip-html-from-strings-in-python. Tested in _sanitize_html"""
+    s = MLStripper()
+    s.feed(html)
+    return s.get_data()
+
+@retry(wait_random_min=20, wait_random_max=30, 
+       stop_max_attempt_number=10)
+def translate(text, translator, chunksize=2000, rate_limit=0.4):
```
russwinch: you didn't use your ratelimit decorator :( 

https://github.com/nestauk/nesta/blob/master/nesta/packages/decorators/ratelimit.py

nesta/production/luigihacks/elasticsearchplus.py
```diff
@@ -19,15 +23,144 @@
...
+        rate_limit (float): Time in seconds to sleep in order to be
+                            kind to Google.
+    Returns:
+        {text, langs} ({str, set}): Translated text and set of 
+                                    detected languages.
+    """
+    time.sleep(rate_limit)
+    chunks = list(sentence_chunks(text, chunksize=chunksize))
+    texts, langs = [], set()
+    for t in translator.translate(chunks):
```
russwinch: I know `dest='en'` is the default language but it would be worth adding it as an argument here for clarity

nesta/production/luigihacks/elasticsearchplus.py
```diff
@@ -403,7 +538,18 @@ def __init__(self, entity_type,
         if caps_to_camel_case:
             self.transforms.append(_caps_to_camel_case)
 
+        # Translate any text to english
+        if auto_translate:
+            urls = list(f"translate.google.{x}"
+                        for x in ('com', 'co.uk', 'co.kr',
```
russwinch: it would be worth clarifying these urls are just for load balancing

nesta/production/orms/meetup_es_config.json
jaklinger: Our priority in the naming convention is clear legibility. Our feeling is that `transl` or even `_translated_` has some ambiguity which disappears with adding the `auto` prefix. The `booleanFlag` is preferred over `bool` since this strictly speaking isn't a type, rather an object. There could be non-boolean flags in the future and we'd prefer to have a book-keeping mechanism for this.

nesta/production/orms/meetup_es_config.json
jaklinger: > * a field like `language_iso2_group` / `language_iso2_organisation` or so (if mentioning the entity type  as the last chunk is not mandatory, I'd use a single identifier like `id_iso2_language`)

A note: because of the implementation, this will be `terms_iso2lang_group` since there could be multiple languages in a piece of text.

jaklinger: we settled on `*_iso*_country` over `*_iso*_<entity>` since our graphs will have countries as entities so, it suits future development to have the suffix as `country`

nesta/production/luigihacks/elasticsearchplus.py
jaklinger: Fine - I put it in :)

nesta/production/luigihacks/elasticsearchplus.py
jaklinger: Yep!

nesta/production/luigihacks/elasticsearchplus.py
jaklinger: Agreed

nesta/production/orms/meetup_es_config.json
mindrones: OK. If we're going in this direction, looking at [aliases](https://github.com/nestauk/nesta/blob/dev/nesta/production/schemas/tier_1/aliases/health_scanner.json) why not do the same for:
- `continent_id` (`id_continent_organisation` -> `id_iso2_continent`)
- `countries_ids` (`terms_of_countryTags` -> `terms_iso2_countryTags`)
- `state_id` (`id_state_*` -> `id_iso2_state`)

Adn btw if we relied on the lookup only these could be removed completely, to reduce the response size:
- `continent`
- `country`
- `state`

Also, as languages could be graph nodes too, why not using `terms_iso2_lang` instead of `terms_iso2lang_group`?

jaklinger: ok, agreed, however if we go along these lines then the best strategy would be to holistically look at everything and rewrite the mappings. 

Opened issue #124

PR #145, opened 2019-08-07
Put the LIVE in arXlive.

russwinch:
This is arXlive!

It's set up to run at 3am GMT Sunday-Thursday (an hour after arXiv update the data, on the days they also update)

- [x] install python-Levenshtein in the container
- [x] investigate why cron cannot find docker-compose
- [x] investigate sql timeout in GRID task
- [x] add 'attempted grid match' flag to prevent re-checking failed matches (cleared when arxiv paper is updated)
- [x] update documentation to cover how this is implemented on the prod machine right now
- [x] add tests to cover new raise when arxiv is up to date


russwinch:
@jaklinger  I'd like to see the tests passing for this so I'll merge dev in here once you have merged your PRs containing the hotfix

docker/luigi_docker.cfg
```diff
@@ -0,0 +1,35 @@
...
+
+[scheduler]
+state_path = /var/lib/luigi/luigi.state
+
+[email]
+format = html
+method = ses
+prefix = [PRODUCTION SYSTEMS]
+# update these to a central email address and then hide this file from git
+receiver =
```
jaklinger: how will `sender` and `receiver` be set?

nesta/packages/arxiv/tests/test_collect_arxiv.py
```diff
@@ -32,188 +31,190 @@ def mock_response():
...
+
+    @mock.patch('nesta.packages.arxiv.collect_arxiv._arxiv_request', autospec=True)
+    def test_arxiv_batch_returns_resumption_cursor(self, mocked_request, mock_response):
+        mocked_request.return_value = ET.fromstring(mock_response)
+        _, reumption_token = arxiv_batch('111222444|1')
+        assert reumption_token == '3132962|1001'
+
+    @mock.patch('nesta.packages.arxiv.collect_arxiv._arxiv_request', autospec=True)
+    def test_arxiv_batch_returns_none_at_end_of_data(self, mocked_request):
+        mock_response = b'''<?xml version="1.0" encoding="UTF-8"?>
```
jaklinger: I would pop this dummy response in a file and retrieve it with a fixture

nesta/packages/arxiv/tests/test_collect_arxiv.py
```diff
@@ -32,188 +31,190 @@ def mock_response():
...
+        total_articles()
+        if 'delay' in mocked_request.call_args[1]:
+            assert mocked_request.call_args[1]['delay'] >= 3
+
+
+class TestArxivBatch:
+    @mock.patch('nesta.packages.arxiv.collect_arxiv._arxiv_request', autospec=True)
+    def test_arxiv_batch_sends_resumption_token_if_provided(self, mocked_request):
+        batch, _ = arxiv_batch('111222444|1')
+        assert mocked_request.mock_calls[0] == mock.call('http://export.arxiv.org/oai2',
```
jaklinger: there are a large number of repeated variables like `http://export.arxiv.org/oai2` and `111222444|1`. Might be an idea to hard code them to some `UPPER_CASE` global (or fixtures, but probs overkill) at the top of the script. (Probably could do this for the `mock.patch` args if they're repeated often

nesta/packages/arxiv/tests/test_collect_arxiv.py
russwinch: as it's only used once I've kept this in as I think it's helpful to see the input with the test, but I was able to halve the amount of xml (there was a lot of redundancy) so now it's comparable to the other blocks in this test file.

nesta/packages/arxiv/tests/test_collect_arxiv.py
russwinch: this is done, looks much better 👍 thanks!

docker/luigi_docker.cfg
russwinch: This is revised. It will pull in the host configurations and therefore set these fields from there. I have removed the docker specific `luigi.cfg`.

There is still a docker specific logging config file, which is passed in as a flag when the task is called.

PR #147, opened 2019-08-09
add general data quality to data_quality branch

dede95:


nesta/packages/dq/general.py
```diff
@@ -0,0 +1,132 @@
...
+    Returns:
+        missing_count_df (:obj:`pandas.core.series.Series`): A series with column names (index) and missing value frequency (column)
+
+
+    """
+    df = pd.DataFrame(input)
+    missing_count_df = df.isnull().sum().sort_values(ascending=False)
+    return missing_count_df
+
+def missing_value_percentage_column_count(input):
```
georgerichardson: I think the functions should probably be standardised to take a dataframe or iterable as inputs

nesta/packages/dq/general.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Args:
+        input (:obj:`dict`): A dictionary of
+
+    Returns:
+        out_counts (:obj:`pandas.core.series.Series`): A series with mising value percentage ranges (index) and column frequency (column)
+
+    """
+
+    missing_bins = [i for i in range(100+1) if i%10 == 0]
```
georgerichardson: Maybe the user should be allowed to specify the bins as a kwarg, with a reasonable default?

nesta/packages/dq/general.py
```diff
@@ -0,0 +1,132 @@
...
+    out= pd.cut((pd.Series(input)/total_length).apply(lambda x: round(x*100,5)),
```
georgerichardson: I think you can take advantage of the fact that you already have a `missing_values` function. How about:

```
m = missing_values(df) / df.shape[0] * 100
m_counts = (pd.cut(m, bins=bins)
            .value_counts(sort=False)
            .reset_index()
            .rename(columns={'index': 'Bins', 0: 'Frequency'}))
return m_counts
```

nesta/packages/dq/general.py
```diff
@@ -0,0 +1,132 @@
...
+                                                         bins=missing_bins, include_lowest=True)
+    out_counts = out.value_counts(sort=False)
+
+    return out_counts
+def missing_value_column_count(input):
```
georgerichardson: This doesn't work as expected when I run it. Maybe this is also redundant. There could just be an arg in the function above that switches between count and percentage

nesta/packages/dq/general.py
```diff
@@ -0,0 +1,132 @@
...
+
+    """
+
+    df = pd.DataFrame(input)
+
+    data_frame_missing = df.isnull().sum() / df.shape[0]
+    out = np.histogram(data_frame_missing, bins=range(0, 101, 10))
+    return out
+
+def missing_value_count_pair_both(input):
```
georgerichardson: This works great. I can't remember what we discussed, but it might be best to return the result as a dataframe or perhaps to have the option to choose between outputting as a dataframe or json

georgerichardson: I didn't look down the page! Looks like you could just have an argument and that can trigger `melt_dict_to_df`

nesta/packages/dq/general.py
```diff
@@ -0,0 +1,132 @@
...
+            either_null_dict[(pair[0],pair[1])] = either_counts.sum()
+            either_null_dict[(pair[1],pair[0])] = either_counts.sum()
+
+        else:
+            either_null_dict[(pair[0],pair[1])] = 0
+            either_null_dict[(pair[1],pair[0])] = 0
+
+    return either_null_dict
+
+def melt_dict_to_df(calc_dict):
```
georgerichardson: This should maybe go in a `utils.py` module as it may well end up being used in other data quality calculations

nesta/packages/dq/general.py
```diff
@@ -0,0 +1,132 @@
+import pandas as pd
+import numpy as np
+import itertools
+
+def missing_values(input):
```
georgerichardson: There could be many more types of null values than those identified by the dataframe. It could be good to have an arg that lets the user specify a list of values that they know to represent missing data.

nesta/packages/dq/general.py
georgerichardson: Add an arg for bins with default as `None` and do
```suggestion
if bins is None:
    bins = range(0, 101, 10)
```

PR #148, opened 2019-08-09
add text data quality and tests to data_quality branch

dede95:


nesta/packages/dq/text.py
```diff
@@ -0,0 +1,79 @@
+import pandas as pd
+import itertools
+from collections import Counter
+##string length
+def string_length(input):#, column):
```
georgerichardson: Maybe include an argument that allows the user to choose whether to count NaN values. If triggered you could do `fillna('')` on the data

nesta/packages/dq/text.py
```diff
@@ -0,0 +1,79 @@
...
+    '''string_counter
+
+    Args:
+        input (:obj:`iter` of :obj:`list`): A sequence of string objects.
+
+    Returns:
+        count (:obj:`Counter`): Counter object.
+    '''
+
+    series = pd.Series(input)
```
georgerichardson: `input` is a built in Python function. Generally best practice not to overwrite them.

nesta/packages/dq/text.py
```diff
@@ -0,0 +1,79 @@
...
+    count = Counter(series)
```
georgerichardson:
```suggestion
    count = (series.value_counts()
                     .reset_index()
                     .rename(columns={'index': 'Text', count.name: 'Frequency'}))
```

nesta/packages/dq/tests/test_text.py
```diff
@@ -0,0 +1,34 @@
...
+from numpy.testing import assert_array_equal
+
+import pytest
+
+# from nesta.packages.dq.text import (string_length, string_counter)
+import sys
+sys.path.insert(0, '/mnt/c/Users/aotubusen/Google Drive/Documents/DS Projects/nesta/nesta/packages/dq')
+from text import (string_length, string_counter)
+
+@pytest.fixture
```
georgerichardson: If you add in the handling of missing values for `string_length` then it might be useful to have another fixture with some missing data too

nesta/packages/dq/tests/test_text.py
```diff
@@ -0,0 +1,43 @@
+import pandas as pd
+import numpy as np
+from collections import Counter
+from numpy.testing import assert_array_equal
+
+import pytest
+
+# from nesta.packages.dq.text import (string_length, string_counter)
+import sys
+sys.path.insert(0, '/mnt/c/Users/aotubusen/Documents/DS Projects/nesta_dq/nesta/nesta/packages/dq/')
```
georgerichardson: Looks like this slipped in. Probably needs to go before merging

PR #149, opened 2019-08-09
add nested data quality and tests to data_quality branch

dede95:


nesta/packages/dq/nested.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Returns:
+        output1 (:obj:`list`): A list of array lengths.
+        output2 (:obj:`Counter`): Counter object of counts.
+
+    '''
+
+    if calculation_type == 'word_length':
+        list_input = [i for i in input if type(i) == list]
+        print(list_input)
```
georgerichardson:
```suggestion
```

nesta/packages/dq/nested.py
```diff
@@ -0,0 +1,56 @@
...
+        output1 = [len(j) for j in itertools.chain(*list_input)]
```
georgerichardson: You don't need `output1` and `output2`. They won't clash because you have the `if` statement

nesta/packages/dq/nested.py
```diff
@@ -0,0 +1,56 @@
...
+        input (:obj:`list`):
+        calculation_type (:obj:`str`): A string indicaing which type of calculation to perform.
+
+    Returns:
+        output1 (:obj:`list`): A list of array lengths.
+        output2 (:obj:`Counter`): Counter object of counts.
+
+    '''
+
+    if calculation_type == 'word_length':
```
georgerichardson: I don't think it's very clear to the user that they would need to give `count` or `word_length` as an argument. First I would change `calculation_type` to a keyword argument and then I would set one of them as a default option. Then I would make sure that the options are well described in the docstring.

nesta/packages/dq/nested.py
```diff
@@ -0,0 +1,56 @@
...
+        list_input = [i for i in input if type(i) == list]
+        print(list_input)
+        output1 = [len(j) for j in itertools.chain(*list_input)]
+        return output1
+
+    elif calculation_type == 'count':
+        list_input = [i for i in input if type(i) == list]
+        output2 = Counter(itertools.chain(*list_input))
```
georgerichardson: Probably best to maintain some consistency with outputs. Although a `Counter` is convenient, I would probably choose to convert it to a `DataFrame` with sensible column names.

nesta/packages/dq/nested.py
```diff
@@ -0,0 +1,56 @@
...
+    Returns:
+        array_length (:obj:`list`): An array of array lengths.
+    '''
+    series = pd.Series(input)
+    # array_length = dataframe[column].apply(lambda x: len(x) if type(x) == type_ else None).values
+    array_length = series.apply(lambda x: len(x) if type(x) == list else None)
+
+    return array_length
+
+def word_arrays(input):
```
georgerichardson: Looks like the code inside this function is duplicated a few times throughout the rest of the module.

georgerichardson:
Some thoughts to bring up in our conversation:
- Can we more elegantly handle string and numeric arrays where possible?
- We should return distributions as well as raw values
- Can we build in `ast.literal_eval` to remove another step from the user

jaklinger:
@dede95 and @georgerichardson there is a risk of this going stale. Better to commit into the upstream feature branch sooner rather than later

PR #172, opened 2019-09-06
Test runner

russwinch:


jaklinger:
will likely migrate to jenkins in daps 2, so this will serve as an example though now deprecated

PR #180, opened 2019-09-12
[157] novelty production

georgerichardson:
Fixed various bugs when populating Project-MeSH term link table and sped up insertion speed by chunking results.

### DIFF
--- a/nesta/core/orms/orm_utils.py
+++ b/nesta/core/orms/orm_utils.py
@@ -21,33 +21,55 @@ import json
 import logging
 import time
 
-def object_to_dict(obj, found=None):
+
+def _get_key_value(obj, key):
+    """Helper method to an attribute value, dealing
+    gracefully with datetimes by converting to isoformat
+
+    Args:
+        obj: Object to retrieve value from
+        key (str): Name of the attribute to retrieve.
+    Returns:
+        {key, value}: A key-value pair corresponding to the attribute
+    """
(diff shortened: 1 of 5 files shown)


nesta/core/orms/orm_utils.py
```diff
@@ -281,7 +281,7 @@ def filter_out_duplicates(db_env, section, database,
             pk = tuple([pkey.type.python_type(row[pkey.name])
                         for pkey in pkey_cols])
             # The row mustn't aleady exist in the input data
-            if pk in all_pks and not is_auto_pkey:
```
jaklinger: i don't think we want to delete this line do we?

nesta/core/routines/health_data/nih_data/mesh_join_task.py
```diff
@@ -81,31 +86,33 @@ def run(self):
...
+            existing_projects = set()
+            projects = session.query(Projects.application_id).distinct()
+            for p in projects:
+                existing_projects.update({int(p.application_id)})
             
             projects_done = set()
             projects_mesh = session.query(ProjectMeshTerms.project_id).distinct()
             for p in projects_mesh:
-                projects_done.update(int(p.project_id))
+                projects_done.update({int(p.project_id)})
```
jaklinger: could do `.add()` rather than `.update({})`

PR #185, opened 2019-09-16
Adding ngrams to arxiv

jaklinger:
First time we're attempting to use n-grams in elasticsearch. Other things in this PR:

- [x] Add preprocessed, tokenised, n-gram'd data to elasticsearch as a terms field
- [x] Make sure the tokens field is searchable
- [x] Clean bad data on ingestion to ES
- [x] Update transnational definition in ES, bringing in line with deep change analysis
- [x] Bring lolvelty online into RootTask using an ES task


jaklinger:
Self-reviewing since this is a minimal addition on top of the `gtr_hotfix` PR

PR #189, opened 2019-09-30
154 clio clusters

### LINKED ISSUE #154: Pipelines to automatically subset documents (clustering) and long query implementation in Clio
- [ ] Clone `dev`, create a new branch with a short sensible name, prefixed with this issue number
- [ ] Create an empty package under `nesta.packages.nlp_utils` called `text2vec`
- [ ] Commit and create a *draft* PR for this branch

** commit regularly after this stage **
[Note: we are using GtR data]

- [ ] Use TTD to write functional code to transform text data into vectors. Where possible, support batching/chunking of data if this leads to performance gains. The output vectors should be written out in chunks on S3.
- [ ] Write an ORM to store the vector groups so that they can be retrieved quickly from the API (later)
- [ ] Use TTD to write functional code which retrieves intermediate text vectors from S3, *yielding* file in their chunks to avoid memory bloat. The code should then generate a cluster model (hard-coded), which can be used to fill the SQL table. The model should be stored on S3.
- [ ] Join it all together with a luigi pipeline. Have a `dev` mode which can run in less than 5/10 minutes. The full `production` mode should assign relatively small clusters to the GtR text field data.

** bonus items **
- [ ] The ORM actually needs to be doubly dynamic (currently it is completely static), so that we can create one lookup table per dataset (without hard-coding), and one table per dataset per date. We've not done this before, but this should help https://stackoverflow.com/questions/19163911/dynamically-setting-tablename-for-sharding-in-sqlalchemy
- [ ] Implement the whole pipeline in autoML. This will require writing a loss function for the clustering stage.
- [ ] Add the 'long query' mode to the AWS lambda function (make a separate PR here https://github.com/nestauk/clio-lite)

kstathou:
Closes #154


jaklinger:
@kstathou what's the status of this- in need of review?

kstathou:
> @kstathou what's the status of this- in need of review?

I wanted to run it on production before asking you for a review. Is it as simple as changing the `test=not self.production` to `test=self.production` in the `RootTask`? Any other considerations?

PR #190, opened 2019-10-01
Ngram fix

jaklinger:
ngrams were excluding any tokens containing terms with fewer than 2 chars, e.g. state-of-the-art

### DIFF
--- a/nesta/packages/nlp_utils/ngrammer.py
+++ b/nesta/packages/nlp_utils/ngrammer.py
@@ -94,7 +94,9 @@ class Ngrammer:
             processed_doc = []
             for sentence in text:
                 sentence = [token for token in sentence
-                            if token not in stop_words]
+                            if token not in stop_words and
+                            not all(len(t) <= 2 
+                                    for t in token.split('_'))]
                 processed_doc.append(sentence)
         return processed_doc
 
--- a/nesta/packages/nlp_utils/ngrammer.py
+++ b/nesta/packages/nlp_utils/ngrammer.py
@@ -95,7 +95,7 @@ class Ngrammer:
             for sentence in text:
                 sentence = [token for token in sentence
                             if token not in stop_words and
-                            not all(len(t) <= 2 
+                            not all(t in stop_words
                                     for t in token.split('_'))]
                 processed_doc.append(sentence)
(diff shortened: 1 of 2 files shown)


nesta/packages/nlp_utils/preprocess.py
```diff
@@ -54,7 +54,7 @@ def clean_and_tokenize(text, remove_stops):
     tokens = tokens_re.findall(text)
     _tokens = [t.lower() for t in tokens]
     filtered_tokens = [token.replace('-', '_') for token in _tokens
-                       if len(token) > 2
+                       if (not remove_stops or len(token) > 2)
```
kstathou: Looks good. My guess is that there are only a few stop words (for example, _of_) that are useful in ngrams. Could you remove those from the `stop_words` instead?

nesta/packages/nlp_utils/ngrammer.py
```diff
@@ -94,7 +94,9 @@ def process_document(self, raw_text, remove_stops=True):
             processed_doc = []
             for sentence in text:
                 sentence = [token for token in sentence
-                            if token not in stop_words]
+                            if token not in stop_words and
+                            not all(t in stop_words
+                                    for t in token.split('_'))]
```
kstathou: What happens with garbage ngrams such as `of_technology` and `the_machine_learning`? Would it be helpful to keep the stop word only when `len(token.split('_')) > 2` or when it is at the middle of the `token.split('_')` list?

nesta/packages/nlp_utils/preprocess.py
jaklinger: actually I did a test and almost every stop word appears in a wiktionary n-gram...

nesta/packages/nlp_utils/ngrammer.py
jaklinger: the assumption is that the number of garbage n-grams is small in the wiktionary dataset. I think this is a limitation at the moment that I can't think of a robust answer for, since terms like `of_biblical_proportions` (actually in the dataset) are relevant, whereas `of_an`  (also in the dataset) are irrelevant, and at least the above procedure works for these examples.

Issue #210, opened 2020-01-15
Arxlive pipeline failure


bishax:
Due to a deprecation in the MAG API (see below), the arxlive pipeline fails in `packages/mag/query_mag_api.py`:

```
Traceback (most recent call last):
  File "/home/ec2-user/miniconda3/lib/python3.6/site-packages/luigi/worker.py", line 199, in run
    new_deps = self._run_get_new_deps()
  File "/home/ec2-user/miniconda3/lib/python3.6/site-packages/luigi/worker.py", line 139, in _run_get_new_deps
    task_gen = self.task.run()
  File "/home/ec2-user/nesta/nesta/core/routines/arxiv/arxiv_mag_task.py", line 186, in run
    update_field_of_study_ids(mag_subscription_key, session, missing_fos_ids)
  File "/home/ec2-user/nesta/nesta/packages/mag/query_mag_api.py", line 201, in update_field_of_study_ids
    in query_fields_of_study(mag_subscription_key, ids=fos_ids)]
  File "/home/ec2-user/nesta/nesta/packages/mag/query_mag_api.py", line 200, in <listcomp>
    new_fos_to_import = [FieldOfStudy(**fos) for fos
  File "/home/ec2-user/nesta/nesta/packages/mag/query_mag_api.py", line 152, in query_fields_of_study
    del row[f]
KeyError: 'E'
ERROR:luigi-interface:[pid 19127] Worker Worker(salt=568461402, workers=10, host=ip-172-31-31-235, username=ec2-user, pid=19099, sudo_user=alex) failed    QueryMagTask(date=2020-01-15, _routine_id=2020-01-15-False, test=True, db_config_env=MYSQLDB, db_config_path=mysqldb.config, mag_config_path=mag.config, insert_batch_size=500, articles_from_date=None)
Traceback (most recent call last):
  File "/home/ec2-user/miniconda3/lib/python3.6/site-packages/luigi/worker.py", line 199, in run
    new_deps = self._run_get_new_deps()
  File "/home/ec2-user/miniconda3/lib/python3.6/site-packages/luigi/worker.py", line 139, in _run_get_new_deps
    task_gen = self.task.run()
  File "/home/ec2-user/nesta/nesta/core/routines/arxiv/arxiv_mag_task.py", line 186, in run
    update_field_of_study_ids(mag_subscription_key, session, missing_fos_ids)
[...]

PR #211, opened 2020-01-15
210 arxlive hotfix

bishax:
Fixes issue #210


bishax:
Requires #213 to be fixed before merging

PR #214, opened 2020-01-22
Add ignore option to find_filepath_from_pathstub

### LINKED ISSUE #212: `find_filepath_from_pathstub` finds docs folder first
First observed in #210 

```
>>> from nesta.core.luigihacks.misctools import find_filepath_from_pathstub as f3p
>>> f3p('batchables/arxiv/arxiv_elasticsearch')
'/home/ec2-user/nesta/docs/_build/html/_modules/nesta/core/batchables/arxiv/arxiv_elasticsearch'
>>> 
```

bishax:
Closes #212

### DIFF
--- a/nesta/core/luigihacks/misctools.py
+++ b/nesta/core/luigihacks/misctools.py
@@ -40,7 +40,7 @@ def get_paths_from_relative(relative=1):
     return map(os.path.abspath, paths)
 
 
-def find_filepath_from_pathstub(path_stub):
+def find_filepath_from_pathstub(path_stub, ignore=('docs/_build',)):
     '''Find the full path of the 'closest' file (or directory) to the current working
     directory ending with :obj:`path_stub`. The `closest` file is determined by
     starting forwards of the current working directory. The algorithm is then repeated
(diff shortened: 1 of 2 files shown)


nesta/core/luigihacks/misctools.py
```diff
@@ -50,14 +50,19 @@ def find_filepath_from_pathstub(path_stub):
 
     Args:
         path_stub (str): The partial file (or directory) path stub to find.
+        ignore(tuple): Tuple of directory sub-strings to ignore
     Returns:
         The full path to the partial file (or directory) path stub.
     '''
+    def is_ignore(path):
+        """ Checks if any `element` of ignore is a substring of `path` """
+        return any([i in path for i in ignore])
```
kstathou: Why did you decide to add a nested function? Wouldn't changing the line 66 to `if path.endswith(path_stub.rstrip("/")) and path not in ignore` achieve the same thing?

bishax: If `ignore` wasn't a tuple then you could do `path.endswith(path_stub.rstrip("/")) and ignore not in path` (note: `ignore not in path` not `path not in ignore`). However, `ignore` is a tuple so that multiple directories can be ignored hence the need for a function that does an `any` check.

I put it inside the body of the function because it is a helper function that isn't needed anywhere else.
Of course, I could've put it outside the function body and called it `_is_ignore` (the leading `_` as a convention says this function shouldn't be used by a user of the package).
This would mean:
- A user could still see and use `_is_ignore` if they wanted to (not so bad really)
- The function would've needed to take an argument `ignore` as `ignore` would be out of scope if `is_ignore` was in the main `.py` body. Then line 66 would read `if path.endswith(path_stub.rstrip("/")) and not is_ignore(path, ignore)` which reads slightly less like English.

kstathou: Great, thanks for the explanation!

PR #223, opened 2020-04-10
Cordis to elasticsearch [prototype for EURITO]

jaklinger:
Refers to issue #222

PR #226, opened 2020-05-01
[224] arxlive lolvelty fix

### LINKED ISSUE #224: [arXlive] fix lolvelty batch size issue
arXlive pipeline fails because of latency in ES lolvelty measuring. Should be able to fix by reducing the batch size.

jaklinger:
Closes #224 

Branched from `222_cordis2es` which is being merged in #223 . Will wait on that PR, before changing the target base to `dev`

In the end a simple fix, relating to the scaling down of the ES cluster. Batch size and number of concurrent requests reduced.


jaklinger:
@bishax 

> Looks good, though is it worth increasing the timeout parameter passed to ElasticsearchPlus.index in batchables/arxiv/arxiv_elasticsearch/run.py to avoid one instance of bad latency crashing a batch job?

I *think* the main point of failure is actually [here] (https://github.com/nestauk/nesta/blob/dev/nesta/packages/novelty/lolvelty.py#L65) since the MLT function is very computationally heavy... but your point still stands.

In the case of the arxlive pipeline, it doesn't really matter since luigi restarts the ES task again because it runs with 10 workers, and so normally has spare capacity to retry the task around 20 times or so before the topic modelling has finished. The restart skips done batch tasks, so the overall efficiency is ok.

I think in general we should avoid increasing the timeout parameter in this particular case because we happen to know that the issue is 99% related to too many concurrent requests, and so the practical solution to the issue is to either pay more and go faster, or pay less and go slower (which is ok)

> Furthermore, does ElasticsearchPlus (or vanilla Elasticsearch) implement any kind of retry behaviour?

Not by default, and there isn't a global setting, but it could be easy enough to force `Elasticsearch` do this by default via `ElasticsearchPlus`. There is a question about whether or not we should do this, since (again) we happen to know that the issue is related to overloading the server, and maybe we wouldn't want to make that easier to do!

PR #230, opened 2020-05-01
[225] biorxiv collection

jaklinger:
Closes #225 

Branched from `arxlive-lolfix` which is being merged in #226. Will wait on that PR, before changing the target base to `dev`.

PR #231, opened 2020-05-05
[227] biorxiv orm

### LINKED ISSUE #227: Amend arxiv table/orms to accept biorxiv
- [ ] Assess fields where this will not be possible
- [ ] Implement change in DB
- [ ] Change ORM to reflect change in DB

jaklinger:
Closes #227 

Branched from `225_biorxiv` which is being merged in #230. Will wait on that PR, before changing the target base to dev.


nesta/core/orms/arxiv_orm.py
```diff
@@ -81,6 +81,7 @@ class Article(Base):
     institutes = relationship('ArticleInstitute')
     corex_topics = relationship('CorExTopic',
                                 secondary='arxiv_article_corex_topics')
+    article_source = Column(VARCHAR(7), index=True, default=None)
```
jaklinger, line 84: Note, I opted for an extra column over a look-up, on the basis that there will be very few values of `article_source` so the subsetting time will be tiny. Also, `VARCHAR(7)` will only add a total of 15MB to the database, which is peanuts.

Issue #228, opened 2020-05-01
Slot biorxiv collection into arxlive pipeline


jaklinger:
- [ ] Add task for data collection
- [ ] Segment the datasets before autoML (`arxiv` only)
- [ ] Duplicate ES tasks for separate `biorxiv` index

PR #235, opened 2020-05-06
[228a] bior/medr/xiv tier-0 collection

jaklinger:
Slight change of plan from #228 (so will not close this issue)

Both biorxiv and medrxiv abstracts can be collected by the same method, so I've collected them both.

The arxlive pipeline had to be amended slightly to allow the filter out biorxiv and medrxiv articles from the upper parts of the pipeline. For now, biorxiv and medrxiv are collected at the bottom of the pipeline (tier-0) into the arxiv database. Issue #228 also specified ES functionality, but that isn't required at this time, and I can park that until required.

I had to increase the PK size to 40 to allow enough freedom to get the jobs through. I'll consider reducing the PK size in the future, but it shouldn't affect performance drastically, rather that it's a bit of a waste of disk space.

Needed to make small extensive changes in response to running on a full dataset, and obviously picking up all the bugs in the process. Also, my deadline was last night, so I needed to prototype quite rapidly.

TODO:
- [x] refactor the collection task into a testable unit
- [x] allow the xiv data to be updated, so that updated citations are present


nesta/core/routines/arxiv/arxiv_collect_iterative_task.py
```diff
@@ -54,15 +54,18 @@ def run(self):
         database = 'dev' if self.test else 'production'
         logging.warning(f"Using {database} database")
         self.engine = get_mysql_engine(self.db_config_env, 'mysqldb', database)
+        Base.metadata.create_all(self.engine)
```
jaklinger, line 57: this needs to be here in the case that you have `DROP`'ped the corresponding tables in `MySQL` to force a clean recollection - i.e. this rebuilds the tables

nesta/core/routines/arxiv/arxiv_collect_iterative_task.py
```diff
@@ -111,8 +114,8 @@ def run(self):
                 count = new_count + existing_count
                 if not count % 1000:
                     logging.info(f"Processed {count} articles")
-                if self.test and count == 1600:
-                    logging.warning("Limiting to 1600 rows while in test mode")
+                if self.test and count == 2000:
```
jaklinger, line 117: Changed the number to something easier to remember - helped with debugging

nesta/core/routines/arxiv/arxiv_grid_task.py
```diff
@@ -171,8 +172,8 @@ def run(self):
             if not count % 100:
                 logging.info(f"{count} processed articles from {total} : {(count / total) * 100:.1f}%")
 
-            if self.test and count == 50:
-                logging.warning("Exiting after 50 articles in test mode")
+            if self.test and count == 5000:
```
jaklinger, line 175: Needed a larger chunk of articles to fully validate that all my institutes were being matched

nesta/core/routines/arxiv/arxiv_grid_task.py
```diff
@@ -189,3 +190,30 @@ def run(self):
         # mark as done
         logging.info("Task complete")
         self.output().touch()
+
+
+class GridRootTask(luigi.WrapperTask):
```
jaklinger, line 195: Added a upstream `RootTask` so I didn't need to run the full pipeline - just the tier-0 collection

nesta/core/routines/arxiv/arxiv_mag_sparql_task.py
```diff
@@ -160,8 +160,8 @@ def run(self):
 
                 if not count % 1000:
                     logging.info(f"{count} done. {total_arxiv_ids_to_process - count} articles left to process")
-                if self.test and count == 150:
-                    logging.warning("Exiting after 150 rows in test mode")
+                if self.test and count == 3000:
```
jaklinger, line 163: Increased to check that all institutes get matched

nesta/core/routines/arxiv/arxiv_mag_task.py
```diff
@@ -189,8 +198,8 @@ def run(self):
                 all_articles_to_update.extend(batch_article_data)
 
                 logging.info(f"Batch {count} done. {total_arxiv_ids_to_process} articles left to process")
-                if self.test and count == 2:
-                    logging.warning("Exiting after 2 batches in test mode")
+                if self.test and count == 10:
```
jaklinger, line 207: Increased number of batches to ensure all institutes are matched

nesta/packages/mag/parse_abstract.py
```diff
@@ -13,8 +13,7 @@ def _uninvert_abstract(inverted_abstract):
     for term, idxs in inverted_abstract['InvertedIndex'].items():
         for idx in idxs:
             terms[idx] = term
-    if None in terms:
```
jaklinger, line 16: this was originally here to catch if such edge cases existed. I found them :p and it turns out that although MAG inverted abstracts have "missing" indexes, from inspection of the "real" articles on e.g. biorxiv that they don't represent anything, i.e. they can be omitted, which is what you will see in the new code

nesta/packages/magrxiv/collect_magrxiv.py
```diff
@@ -0,0 +1,45 @@
+from nesta.packages.mag.query_mag_api import get_journal_articles
```
jaklinger, line 1: this is actually the same as the `biorxiv_collect` that you previously reviewed, but I've renamed to reflect the fact that it now picks up bio and medrxiv articles

nesta/core/orms/arxiv_orm.py
```diff
@@ -22,7 +22,7 @@
 """Association table for Arxiv articles and their categories."""
 article_categories = Table('arxiv_article_categories', Base.metadata,
                            Column('article_id',
-                                  VARCHAR(20),
+                                  VARCHAR(40),
```
jaklinger, line 25: The new size is briefly justified in the PR initial comment...

nesta/core/routines/arxiv/arxiv_mag_task.py
```diff
@@ -62,6 +63,14 @@ def requires(self):
                        test=self.test,
                        articles_from_date=self.articles_from_date,
                        insert_batch_size=self.insert_batch_size)
+        yield CollectMagrxivTask(date=self.date,
+                                 routine_id=self._routine_id,
+                                 db_config_path=self.db_config_path,
+                                 db_config_env=self.db_config_env,
+                                 test=self.test,
+                                 articles_from_date='1 April 2020' if self.test else '1 January 2010',
```
bishax: `'1 April 2020'` should be changed to something dependent on the current date to maintain short pipeline tests in the future.

nesta/core/routines/arxiv/magrxiv_collect_iterative_task.py
```diff
@@ -0,0 +1,120 @@
...
+    Args:
+        date (datetime): Datetime used to label the outputs
+        db_config_env (str): environmental variable pointing to the db config file
+        db_config_path (str): The output database configuration
+        insert_batch_size (int): Number of records to insert into the database at once
+        articles_from_date (str): Earliest possible date considered to collect articles.
+    '''
+    date = luigi.DateParameter()
+    routine_id = luigi.Parameter()
+    test = luigi.BoolParameter(default=True)
```
bishax, line 107: should this be `production` (and `False`) to stick to convention?

nesta/core/routines/arxiv/magrxiv_collect_iterative_task.py
```diff
@@ -0,0 +1,120 @@
...
+                for row in get_magrxiv_articles(xiv, api_key, start_date=articles_from_date):
+                    if row['id'] in all_article_ids:
+                        continue
+                    all_article_ids.add(row['id'])
+                    articles.append(row)
+                    if not len(articles) % 1000:
+                        logging.info(f"Processed {len(articles)} articles")
+                        if self.test:
+                            break
+                if self.test and len(articles) > 1000:
```
bishax: Is this meant to stop at `1000` or `2000`? I.e. should it be `>=`?
(Above line breaks *at* `1000`)

nesta/core/routines/arxiv/magrxiv_collect_iterative_task.py
```diff
@@ -0,0 +1,120 @@
...
+                    logging.warning(f"Limiting to {len(articles)} rows while in test mode")
+                    break
+                logging.info(f"Retrieved {len(articles)} articles so far")
+                    
+            # insert any remaining new and existing articles            
+            logging.info("Inserting data")    
+            print(len(articles), len(set((row['id'] for row in articles))))
```
bishax: log/delete?

nesta/core/routines/arxiv/magrxiv_collect_iterative_task.py
```diff
@@ -0,0 +1,120 @@
...
+            objs = insert_data(self.db_config_env, "mysqldb", database,
+                               Base, Article, articles, low_memory=True)
+            logging.info(f"\t\tInserted {len(objs)}")
+
+        # mark as done
+        logging.warning("Task complete")
```
bishax: warning?

nesta/core/routines/arxiv/magrxiv_collect_iterative_task.py
```diff
@@ -0,0 +1,120 @@
...
+    
+    def requires(self):
+        routine_id = "{}-{}".format(self.date, self.production)
+        logging.getLogger().setLevel(logging.INFO)
+        yield CollectMagrxivTask(date=self.date,
+                                 routine_id=routine_id,
+                                 test=not self.production,
+                                 db_config_env='MYSQLDB',
+                                 db_config_path=self.db_config_path,
+                                 articles_from_date='1 April 2020')
```
bishax: Similar issue to previously with date

nesta/packages/mag/parse_abstract.py
```diff
@@ -13,8 +13,7 @@ def _uninvert_abstract(inverted_abstract):
...
-        raise ValueError('Inverted abstract is incomplete')
+    terms = list(filter((None).__ne__, terms))  # Filter missing terms
```
bishax, line 16: Nifty filter

jaklinger:
Yeah, was going to tick the check boxes at the bottom first. Will be ready
for review tomorrow morning!

bishax:
I won't be able to review until Monday but as long as the test pipeline runs then only my comments are the minor ones above (so no need to wait as long as test pipeline passes)

nesta/core/routines/arxiv/magrxiv_collect_iterative_task.py
jaklinger: ah, so actually the `>` ensures that it will do at least one more loop in test mode, to pick up multiple `xivs`. I'll add a comment to clarify that

jaklinger:
> Not yet run test pipelines, only read through.
> Saw you removed request to review?

Yes I'll implement some refactoring/slight modification as per the two bullet points in the first comment. Likely that I won't get around to doing this til Monday or Tuesday anyway!

nesta/core/routines/arxiv/arxiv_mag_task.py
jaklinger: Good point

nesta/core/routines/arxiv/magrxiv_collect_iterative_task.py
jaklinger: the convention is `production` for RootTasks (since it should be off by default and therefore explicitly implied from the command line), and `test` for all other tasks (for syntax, as it is short and generally the "exception" case)

nesta/core/routines/arxiv/magrxiv_collect_iterative_task.py
```diff
@@ -15,6 +16,74 @@
...
+        flush_every (int): Sets the number of articles retrieved between log messages,
+                           and also related to the maximum number of articles acquired
+                           in test mode (NB: in test mode at most 2 x flush_every will be collected).
+    Returns:
+        articles (:obj:`list` of :obj:`dict`): List of articles from MAG, formatted a-la-arxiv.
+    """
+    logging.info(f"{len(article_ids)} existing articles will not be recollected")
+    articles = []
+    for xiv in xivs:
+        articles += _get_articles(xiv, api_key, start_date, article_ids, test, flush_every)
```
bishax: Rather than hidden mutation of `article_ids` could do something along the lines of the following:

```python
# 1
import operator as op
article_ids.update([op.getitem(row, 'id') for row in articles])
# 2
from toolz.curried import get_in
article_ids.update(map(get_in('id'), articles)
```

`_get_articles` could then be a comprehension (if it wasn't for the logging)

The logging could actually just be a re-usable generator function that wraps an iterable, something like:
 ```python

def foo(x, flush_every, test=True): 
     counter = 0 
     x = iter(x) 
     for xi in x: 
         counter += 1 
         if counter % flush_every == 0: 
             logging.info(f'Processed {counter} items') 
             if test:
                 return xi 
         yield xi 

# in body of `get_articles`:
if test:
    _get_articles = foo(_get_articles)
for xiv in xivs:
    articles = [row for row in get_magrxiv_articles(xiv, api_key, start_date=date) if row['id'] not in article_ids]
    article_ids.update(map(get_in('id'), articles)
```

File: nesta/core/routines/arxiv/magrxiv_collect_iterative_task.py [hunk 18]
```diff
@@ -45,48 +116,20 @@ def output(self):
...
-            logging.info(f"\t\tInserted {len(objs)}")
-
-        # mark as done
-        logging.warning("Task complete")
+        # Database setup and connection
+        database = 'dev' if self.test else 'production'
+        logging.info(f"Using {database} database")
+        engine = get_mysql_engine(self.db_config_env, 'mysqldb', database)
+        with db_session(engine) as session:
+            article_ids = set() if not self.dont_recollect else get_article_ids(session, self.xivs)
```
bishax, line 131: this top-level is a lot more readable :slightly_smiling_face:

nesta/core/routines/arxiv/magrxiv_collect_iterative_task.py
bishax: Actually, this solution doesn't work :facepalm: the `article_ids` definition is one level above so there's still hidden mutation. Would need to take a copy of the original input too.

jaklinger: Yep, good shout. Creating a copy of `article_ids` isn't a big issue, and then the ids can just be updated as you suggest.

PR #237, opened 2020-05-10
[236] Add inf exception and reduced server load

### LINKED ISSUE #236: GtR infinite value of fundingId
Somehow GtR API has started to return a couple`fundingId` with a value of `inf`. From inspection, these look like they should be treated as `null`, and so the `inf` has been added to the list of terms to be considered as `null`.

jaklinger:
Closes #236 

- [x] GtR server load reduced by scaling down batch parameters, to improve the stability of the data collection
- [x] Added an edge case exception for the value `inf` as a null value (see #236)


georgerichardson (approved):
Looks good. Tests pass and runs in test mode on production server. Good to merge.

PR #246, opened 2020-05-19
[243] Make sure all text fields have a `keyword` field to enable aggregations.

mindrones:
Closes #243

PR #251, opened 2020-05-21
[250] VARCHAR length wasn't in pkey check

jaklinger:
When doing `insert_batch`, `VARCHAR` lengths greater than truncated keys weren't considered, leading to duplicate key errors.


jaklinger:
Note, this will be merged into #246 in order to get all the pipelines run, but won't be included in the code review/PR, as I plan to merge this first anyway.

Issue #215, opened 2020-01-22
Create unified ES end-point for health scanner


bishax:
This would unify the three datasets under a common schema, alleviating the need for GraphQL whose cold lambda starts are currently causing poor performance.
This will be implemented as a Luigi task that will take data from the latest ES indices for each dataset, apply a remapping and insert them under a unified index (adding a flag for the dataset type).

- [ ] Write schema remappings
- [ ] Write tasks to apply remappings
  - [ ] NiH
  - [ ] Crunchbase
  - [ ] Meetup

mindrones:
> adding a flag for the dataset type

there should be [`type_of_entity`](https://github.com/nestauk/imt_playbook/blob/dev/schema/nih/mosaic_nih_v5.yaml#L41)

> Write schema remappings

I'll do this now for the discussion

mindrones:
This is just comparing CB, MU, NIH for Mosaic (not EURITO).

### unused

```yaml
cb:
  _cost_usd2018_organisation: float
  _terms_sdg_summary: string[]

mu:
  _id_state_group: string
  _placeName_state_group: string
  _terms_memberOrigin_group: string[]
  _terms_sdg_description: string[]
```

### identification

```yaml
id:
  mu:
    id_of_group: string
  nih:
    id_of_project: string
name:
  cb:
    name_of_organisation: string
  mu:
    name_of_group: string
  nih:
    title_of_organisation: string
```

### content

```yaml
description:
  cb:
    textBody_descriptive_organisation: string
  mu:
    textBody_descriptive_group: string
  nih:
    textBody_descriptive_project: string
brief:
  cb:
    textBody_summary_organisation: string
  nih:
    textBody_abstract_project: string
title:
  nih:
    title_of_project: string
```

### time

```yaml
start_date:
  cb:
    date_birth_organisation: date # yyyy-MM-dd
  mu:
    date_start_group: date # yyyy-MM-dd
  nih:
    date_start_project: date # yyyy-MM-dd
end_date:
  cb:
    date_death_organisation: date # yyyy-MM-dd
  nih:
    date_end_project: date # yyyy-MM-dd
update_date:
  cb:
    datetime_updated_organisation: date # yyyy-MM-dd HH:mm:ss
```

### geo

```yaml
continentName:
  cb, nih:
    placeName_continent_organisation: string
  mu:
    placeName_continent_group: string
continentId:
  cb, nih:
    id_of_continent: string
    id_continent_organisation: string
  mu:
[...]

PR #281, opened 2020-06-03
[267] Tidy & slim schema transformations

jaklinger:
Related to #267, which isn't closed until #280 is addressed (which is the base of this branch).

- [x] Prune deprecated schema transformations
- [x] Only one ontology per dataset (no project specific ontologies)
- [x] Tidy directory naming, and refactor references accordingly
- [x] Add `entity_type` to new dataset ontologies
- [x] Remove unused ontologies from `tier_1.json` (renamed `ontology.json`)
- [x] Add tests to ensure ES mappings are both self-consistent and consistent with the slimmed schema transformations
- [x] Rollback any inadvertently changed json files, unrelated to this PR
- [x] Confirm all new mappings and old mappings are equivalent, then prune old mappings
- [x] Spawn PR #282  to iterate on `dev` pipeline validation


jaklinger:
@mindrones 
For the review, could you please consider any file under `nesta/core/schemas/tier_1/`. All other changes are in the python code to reflect corresponding simplifications in the directory and config structure.

nesta/core/schemas/tier_1/datasets/arxiv.json
```diff
@@ -0,0 +1,28 @@
+{
+    "entity_type": "article",
```
mindrones, line 2: I would prefer if we called this `source` as I'm suggesting [here](https://github.com/nestauk/nesta/issues/215#issuecomment-636059532) so to reserve the "entity" terminology for actual entities (City, Country, Person, Org, etc) to introduce later on when we'll start using ontologies.

This would also make it easier to have documents from different `source`s in the same index, e.g. `arxlive` having `source` = `arxiv` | `biorxiv` | `medrxiv`.

nesta/core/schemas/tier_1/datasets/arxiv.json
```diff
@@ -0,0 +1,28 @@
...
+    "tier0_to_tier1": {
```
mindrones, line 3: Here I'd use a symbol as a separator like `tier0,tier1` or `tier0:tier1`.

nesta/core/schemas/tier_1/mappings/defaults/defaults.json
```diff
@@ -1,24 +1,24 @@
 {
-    "mappings":{
-	"_doc":{
```
mindrones: (just a comment, in Svizzle I'm actually going back to tabs as it's editor configurable so it makes everyone happy :)

mindrones (commented):
Just noted down a couple of changes I'd like to see, but not really sure what implications these would have.

nesta/core/schemas/tier_1/datasets/arxiv.json
jaklinger: I wouldn't take this field too seriously, it will be removed in daps2 as the entity and source will be tied to fields rather than the entire dataset - and these will be automatically propagated. It's basically in here as a legacy support to HealthMosaic which required an general entity type for each dataset.

mindrones: OK

nesta/core/schemas/tier_1/datasets/arxiv.json
jaklinger: are you suggesting that I should use a comma or colon in a field name? 😲again, this will also removed in daps2 (in favour of implicit mapping of raw field names to standardised names) so it won't burden us for too long.

nesta/core/schemas/tier_1/mappings/defaults/defaults.json
jaklinger: oops, actually I have a new unit test that this would have failed (all json in the repo must be clean to pass) - so this will fail anyway :)

nesta/core/schemas/tier_1/datasets/arxiv.json
mindrones: yes, :) I'd favour `tier0:tier1`
OK.

mindrones (approved):
Good to merge for my part 👍

PR #301, opened 2020-07-28
[300] Crunchbase to ES

jaklinger:
Partially addresses #300, which will be closed in the next PR #302

This PR deals with curation of Crunchbase data, prior to transfer to the `general` endpoint. Next PR will transfer to the `general` endpoint. This is a template for refactoring all production pipelines to enforce:

- a separation of dataset-specific collection and enrichment from the project-specific cleaning and curation
- conistency between data in the data warehouse (MySQL) and data products (e.g. Elasticsearch) 

You can run the test for this pipeline using (from `nesta/core/routines/projects/general`):

`luigi --module general_curate CurateTask`

Issue #264, opened 2020-05-29
[DAPS, metaissue] Create integrated pipeline for data collections to ES


jaklinger:
- [x] #265 Add versioning for config management
- [x] #266 Migrate to base ES config, to remove repetition from config
- [x] #267 Migrate to template ES mappings, to remove repetition from mappings
~~- [ ] #268 Create Sql2ES subclass to pick up pipeline dependencies~~
- [x] #269 Add new `general` endpoint (ES7)
- [x] #270 (`general`) Create GtR mapping and validate mapping for ES7
- [x] #271 (`general`) Request validation on GtR data, then run in production mode

Prerequisite for the following: 
- [x] #240 Migrate `arxlive`  
- [x] #241 Migrate `eurito`
- [x] #199 fix CB collection pipeline, due to new API version

for each (~~`gtr`~~, ~~`arxiv`~~, ~~`companies`~~, ~~`nih`~~, ~~`patstat`~~ #313, ~~`cordis`~~ #315)
- [x] (`general`) Create new mapping for ES7 + validate mapping
- [x] (`general`) Request validation on data, then run in production mode, add collect
- [x] Add all `general` pipelines to weekly schedule

Regarding endpoint migration from ES6 to ES7:
~~- [ ] Run AWS's migration from `ES6.x` to `ES7.x` on `eurito-dev`, and validate~~
~~- [ ] Run AWS's migration from `ES6.x` to `ES7.x` on `health-scanner`, and validate~~
- [x] Run AWS's migration from `ES6.x` to `ES7.x` on `arxlive`, and validate

finally
- [x] Rearrange pipelines into datasets and projects
- [x] provide training for Luca & Seb

jaklinger:
Items which in the end will not be addressed in this issue:

- NiH data is not added to the `general` pipeline yet, since the deduplication strategy currently relies on creating two indexes in Elasticsearch, {one for all documents including dupes and applying deduplication logic}, and {one for the deduped documents}. This doesn't fit within the `general` pipeline paradigm and will require significant rewiring to make this possible (of course, it is doable, but starts to fall out of scope for now). Created issue #317
- `eurito-dev` is deprecated in favour of `general`, and so will not be upgraded to ES7
- `health-scanner` is deprecated in favour of `health-mosaic`, and will be dealt with when I'm given the green light to work on that.

PR #302, opened 2020-07-30
[300] Crunchbase to general es pipeline

### LINKED ISSUE #300: Crunchbase to general endpoint
Lightweight transfer of crunchbase data to the `general` endpoint

jaklinger:
Closes #300

Transfer of curated crunchbase data (see #301) to `general` endpoint.

Run with `luigi --module general_root RootTask`

from `nesta/core/routines/projects/general/`

ES endpoint is `general`, details can be found in `elasticsearch.yaml`


nesta/core/schemas/tier_1/tests/test_ontology.py
```diff
@@ -21,7 +21,7 @@ def test_ontology_uniqueness(ontology):
 def test_validate(ontology):
     dirname = Path(os.path.dirname(__file__)).parent
     dataset_dirname = os.path.join(dirname, 'datasets')
-    firsts, middles, lasts = [], [], []
+    firsts, middles, lasts = set(), set(), set()
```
jaklinger, line 24: sets instead of lists, since the assert error message is easier to read when duplicates aren't in

nesta/core/scripts/docker_recipes/py37_amzn1_es7/Dockerfile
```diff
@@ -17,14 +17,15 @@ RUN yum install which -y
 RUN yum install findutils -y
 RUN yum install wget -y
 RUN yum install gcc -y
+RUN yum install geos-devel -y
```
jaklinger, line 20: shapely in py37 build fails without this

nesta/core/routines/projects/general/general_curate.py
```diff
@@ -56,7 +56,7 @@ def requires(self):
                               memory=2048,
                               intermediate_bucket=S3_BUCKET)
 
-        params = (('crunchbase', CrunchbaseOrg.id),)
+        params = (('companies', CrunchbaseOrg.id),)
```
bishax, line 59: Is `companies` too general? E.g. what happens if (when?) we add glass data?

nesta/core/luigihacks/sql2estask.py
```diff
@@ -48,8 +49,8 @@ class Sql2EsTask(autobatch.AutoBatchTask):
     drop_and_recreate = luigi.BoolParameter(default=False)
     dataset = luigi.Parameter()
     endpoint = luigi.Parameter()
-    id_field = luigi.Parameter()
-    filter = luigi.Parameter(default=None)
+    id_field = SqlAlchemyParameter()
+    filter = SqlAlchemyParameter(default=None)
     entity_type = luigi.Parameter()
     kwargs = luigi.DictParameter(default={})
```
bishax, line 56: Note: L97 references organisations but this should be generic?

bishax (commented):
Generally looking good, though I'm confused on how PR's 301 and 302 tie together?
Also, when testing my Luigi never reported any kind of `:)`/`:(` termination message (and I can't see a process running anywhere)

nesta/core/routines/projects/general/general_curate.py
jaklinger: this is a contractual obligation, that for outward facing products we shouldn't refer to CB by name for individual records (i.e. only to credit as a data source), so at some point in the pipeline we need to switch from CB to something else. In the health scanner, the choice was `companies`. As for glass, this is a good question - and perhaps we will use the same entity name `companies` and accept different coverage, or we can revisit the design choices. Good question!

jaklinger:
> Generally looking good, though I'm confused on how PR's 301 and 302 tie together?

@bishax fair point. None of 290/301/302 actually tie together formally *yet*, since one of the objectives of #264 is to decouple `datasets` and `projects`, and then building two sets of pipelines:

- individual pipelines for dataset collection (these already exist)
- one unified `general` pipeline for curating + piping2es all data (with switches for individual datasets), which don't rely on rerunning the full data collection.

This second point is to allow front-end developers to make changes to schemas, data quality issues in curation, etc.

> Also, when testing my Luigi never reported any kind of `:)`/`:(` termination message (and I can't see a process running anywhere)

Yeah I also noticed that. I think it's because I rediscovered the `set_log_level` util in the codebase, and it's something in there that's making this happen. I'll resolve that this week with issue: #303

PR #304, opened 2020-08-06
[293] Text2Vector with luigi

### LINKED ISSUE #293: [AI Diversity] Transform text to vector
Transform variable-length text (usually ranging from a few sentences to a paragraph) to a dense vector using [sentence transformers](https://github.com/UKPLab/sentence-transformers). 

This will be a metaflow task.

### LINKED ISSUE #303: [DAPS] Where has the luigi smiley face gone?
`set_log_level` is removing the `luigi` CLI smiley report. Get it back.

jaklinger:
Closes #293 
Closes #303 

Convert text to vectors, via BERT in batches with luigi. Note, I've written a very generic module so that we can convert any text field in the database to vectors, which are then stored in MySQL JSON format.

A subsequent PR addresses the soft clustering described in #293, which will also be generic.


nesta/core/luigihacks/misctools.py
```diff
@@ -61,3 +64,64 @@ def find_filepath_from_pathstub(path_stub):
...
                 return path
         relative += 1
+
+
+def f3p(path_stub):
+    """Shortened name for coding convenience"""
+    return find_filepath_from_pathstub(path_stub)
+
+
+def load_yaml_from_pathstub(pathstub, filename):
```
jaklinger, line 74: note: migrated here from `orm_utils` since it's not an ORM util and it's only used in `orm_utils` and `misctools`, so migration was straightforward

PR #306, opened 2020-08-13
[305] update GRID lookup logic

### LINKED ISSUE #305: [DAPS] GRID alias lookup was unaware of aliases in the main table
The current implementation of the GRID alias lookup ignores the fact that there are a small number of "alias" institutes which GRID has marked as separate entities.

For example `University of Korea` has two entries in the main GRID table, one based in Seoul and the other in Tokyo. Clearly, the proposed implementation (upcoming PR) will be incorrect in the case where there are two coincidentally similar names. I'm hesitant to implement a fix for that case right now, as it will require rewiring in the arXlive (and other?) pipeline(s) and so is a bit less light-touch.

The downstream effect for arXlive is that (e.g.) `University of Korea` will be marked as a transnational organisation, whereas it wasn't previously. In the upcoming CB-GRID matching PR, this is absolutely required to match organisations based on country and name (`University of Korea` is a concrete example).

jaklinger:
Closes #305

### DIFF
--- a/nesta/packages/grid/grid.py
+++ b/nesta/packages/grid/grid.py
@@ -141,19 +141,22 @@ def grid_name_lookup(engine):
     Returns:
         (:obj:`list` of :obj:`dict`): lookup table [{name: [id1, id2, id3]}]
                 Where ids are different country entities for multinational institutes.
-                Most entities just have a singe [id1]
+                Most entities just have a single [id1]
     """
+    institute_name_id_lookup = defaultdict(set)
     with db_session(engine) as session:
-        institute_name_id_lookup = {institute.name.lower(): [institute.id]
-                                    for institute in session.query(Institute).all()}
+        for institute in session.query(Institute).all():
+            name = institute.name.lower()
+            institute_name_id_lookup[name].add(institute.id)
         logging.info(f"{len(institute_name_id_lookup)} institutes in GRID")
 
         for alias in session.query(Alias).all():
-            institute_name_id_lookup.update({alias.alias.lower(): [alias.grid_id]})
+            name = alias.alias.lower()
+            institute_name_id_lookup[name].add(alias.grid_id)
         logging.info(f"{len(institute_name_id_lookup)} institutes after adding aliases")
 
         # look for institute names containing brackets: IBM (United Kingdom)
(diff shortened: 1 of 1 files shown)


nesta/packages/grid/grid.py
```diff
@@ -141,33 +141,37 @@ def grid_name_lookup(engine):
...
         # look for institute names containing brackets: IBM (United Kingdom)
-        with_country = defaultdict(list)
+        n_countries = 0
         for bracketed in (session
                           .query(Institute)
                           .filter(Institute.name.contains('(') & Institute.name.contains(')'))
                           .all()):
             found = re.match(r'(.*) \((.*)\)', bracketed.name)
             if found:
                 # combine all matches to a cleaned and lowered country name {IBM : [grid_id1, grid_id2]}
```
bishax: - `s/country name/institute name`?
- `{ibm: [grid_id1, grid_id2]}`?

nesta/packages/grid/grid.py
```diff
@@ -141,33 +141,37 @@ def grid_name_lookup(engine):
...
-        logging.info(f"{len(with_country)} institutes with country name in the title")
-
-        # append cleaned names to the lookup table
-        institute_name_id_lookup.update(with_country)
-        logging.info(f"{len(institute_name_id_lookup)} total institute names in lookup")
-
+                name = found.groups()[0].lower()
+                institute_name_id_lookup[name].add(bracketed.id)
+                n_countries += 1
+        logging.info(f"{n_countries} institutes with country name in the title")
```
bishax, line 170: Not familiar with the data. Is it guaranteed that all bracketed terms reference countries?

bishax (commented):
Not run any code but other than the 2 comments then looks good

nesta/packages/grid/grid.py
jaklinger: It was guaranteed when the original code was written (Russ did a side check to see whether this was true) but GRID is always adding institutes, so in principle this could not be true (at present or in the future). In **hypothetical** edge cases such as:

- Microsoft (Teams)
- Microsoft (Comms)

we would anyway want to combine them in the same way, but I haven't seen any examples of this in the GRID data thus far.

PR #327, opened 2020-10-12
[326] Refactoring and tidying of NiH collection

### LINKED ISSUE #51: Bad data quality in some NiH
For example, search for "Project? ?Summary/Abstract This? ?SBIR? ?Phase? ?I?"

Diagnostic shows around 0.0004% of results in dataset contain at least one phrase like "? ?".

### LINKED ISSUE #326: [EURITO] Factor out dataset-specific processing from NiH
In line with the newer "dataset <--> project" separation paradigm for data processing, the NiH dataset-specific collection/processing should be factored out of the Health Mosaic pipeline, with the Health Mosaic-specific pipelines to be considered again later, if required.

This will allow for the data to be immediately useable to address EURITO/Pivot#32

Tasks:

- [x] Collect NiH with refactored pipeline, whilst fixing #51 
- [ ] Deduplicate NiH data without elasticsearch
- [ ] Select dedupes, aggregate relevant fields and add enrichment fields (such as countries) - all into one unified table
- [ ] Setup data_getter

jaklinger:
Addresses (but doesn't close) #326 
Closes #51 

This is the first in a string of PRs for refactoring NiH collection and processing pipelines to be non-project specific. 

Tasks for this PR:

- [x] Backup the old database
- [x] Drop all NiH tables
- [x] Collect and clean (as much as possible) NiH data

Cleaning steps:

- [x] Split `;` terms into an array 
- [x] CAPS --> Camel Case
- [x] Address the bad dq issue highlighted in #51 
~~- [ ] Check Greek characters (etc) parse ok~~ (see comment below)
- [x] Check question marks (bad unicode parsing) fixed
- [x] "NULL", "", "N/A", `[]` --> `null`

You can test this pipeline with:

```python
luigi --module nih_collect_task RootTask
```

Future PRs will:

1. Geocode and apply country-tagging as a lookup table
2. Implement a generic highly scalable deduplication strategy (including very close matches), which can be ported to other datasets. Output is a lookup table.
3. under data-getters, will write a getter for analysis of deduped projects


jaklinger:
Note: "bad greek character" encoding (which is already in the raw data, nothing to do with us) can only reasonably achieved with data science to infer values. The cost-benefit of this is likely quite shaky.

e.g.

The "?" in "?-synuclein" could be 'alpha', 'beta' or 'gamma'. Without apply some kind of likelihood based metric, we will be inferring the wrong character.

There are other data sciency data cleaning tasks that could be performed, such as joining together terms separated by hyphens, if they correctly match context. This would be to fix line breaks (e.g. hell- o world).

Neither of the above will be fixed in this PR, it's simply too time consuming and likely to achieve ok results at best.

nesta/core/luigihacks/autobatch.py
```diff
@@ -92,7 +93,9 @@ class AutoBatchTask(luigi.Task, ABC):
     test = luigi.BoolParameter(default=True)
     max_live_jobs = luigi.IntParameter(default=25)
     worker_timeout = float('inf')
-    
+    date = luigi.DateParameter(default=dt.now())
```
jaklinger, line 96: very minor addition - just means this is in here by default to avoid boilerplate. there can literally be no side-effect of this, promise!

PR #328, opened 2020-10-26
[326 precursor] File/variable renaming factored out from 326_nih

jaklinger:
Refers to #326

Refactors the name `health_data` to `nih` across the system.

The only addition on top of this are 3 dummy routines which will be filled in child PRs.


nesta/core/orms/mesh_orm.py
```diff
@@ -22,11 +22,8 @@ class MeshTerms(Base):
 class ProjectMeshTerms(Base):
     """Association table for NIH projects and their MeSH terms."""
     __tablename__ = 'nih_mesh_terms'
-    project_id = Column(
-            INTEGER,
-            ForeignKey(Projects.application_id),
```
bishax, line 27: Is the ForeignKey intented to be dropped?

nesta/core/orms/nih_orm.py
```diff
@@ -2,116 +2,122 @@
...
+from sqlalchemy.types import INTEGER, JSON, DATETIME
 from sqlalchemy import Column, Table
 
 
 Base = declarative_base()
+TEXT = _TEXT(collation='utf8mb4_unicode_ci')
+VARCHAR = lambda *args, **kwargs: _VARCHAR(*args, 
+                                           collation='utf8mb4_unicode_ci', 
+                                           **kwargs)
+
```
bishax: Minor nitpick: a partial would be clearer

nesta/core/orms/nih_orm.py
```diff
@@ -2,116 +2,122 @@
...
 
 class Projects(Base):
     __tablename__ = 'nih_projects'
 
-    application_id = Column(INTEGER, primary_key=True)
+    application_id = Column(INTEGER, primary_key=True, autoincrement=False)
```
bishax, line 24: Why is the autoincrement dropped?

nesta/core/orms/nih_orm.py
```diff
@@ -2,116 +2,122 @@
...
     org_district = Column(INTEGER)
-    org_duns = Column(VARCHAR(75))
-    org_fips = Column(VARCHAR(2))
+    org_duns = Column(JSON)
+    org_fips = Column(VARCHAR(2), index=True)
     org_ipf_code = Column(INTEGER)
-    org_name = Column(VARCHAR(92))
-    org_state = Column(VARCHAR(2))
+    org_name = Column(VARCHAR(100), index=True)
+    org_state = Column(VARCHAR(2), index=True)
```
bishax, line 49: A lot of added indexes - what are the performance implications? Asking for a friend :sweat_smile:

bishax: When to use `VARCHAR` vs `TEXT`?

nesta/core/orms/nih_orm.py
```diff
@@ -2,116 +2,122 @@
...
     affiliation = Column(TEXT)
-    author_list = Column(TEXT)
-    country = Column(VARCHAR(25))
+    author_list = Column(JSON)
+    country = Column(VARCHAR(50), index=True)
     issn = Column(VARCHAR(9))
     journal_issue = Column(VARCHAR(75))
-    journal_title = Column(VARCHAR(282))
-    journal_title_abbr = Column(VARCHAR(108))
+    journal_title = Column(VARCHAR(400), index=True)
```
bishax, line 89: 400 char journal name? :scream:

nesta/core/orms/nih_orm.py
```diff
@@ -2,116 +2,122 @@
...
+
 
-# class LinkTables(Base):
-#     __tablename__ = 'nih_linktables'
+class LinkTables(Base):
+    __tablename__ = 'nih_linktables'
 
-#     pmid = Column(INTEGER, primary_key=True)
-#     project_number = Column(VARCHAR(11), index=True)
+    pmid = Column(INTEGER, primary_key=True, autoincrement=False)
```
bishax, line 112: pm is pubmed?

**Point for DAPS2:** ORMs alone are not sufficient, we need a process for generating (human-readable) documentation that stays in sync with the ORMs

nesta/core/orms/orm_utils.py
```diff
@@ -428,14 +453,116 @@ def _filter_out_duplicates(session, Base, _class, data,
         if not is_auto_pkey and not low_memory and session.query(exists(_class, **row)).scalar():
             existing_objs.append(row)
             continue
-        objs.append(_class(**row))
+        #objs.append(_class(**row))
+        objs.append(row)
```
bishax: Return type changed from ORM object to dict?

nesta/core/luigihacks/misctools.py
```diff
@@ -125,3 +125,18 @@ def extract_task_info(luigi_task):
     task_name = type(luigi_task).__name__
     routine_id = f'{task_name}-{luigi_task.date}-{test}'
     return test, routine_id
+
+
+@lru_cache()
+def bucket_keys(bucket_name):
```
bishax: Where does this get used?

nesta/core/orms/orm_utils.py
```diff
@@ -428,14 +453,116 @@ def _filter_out_duplicates(session, Base, _class, data,
...
+        for col in rows[0].keys():
+            value = None
+            for row in rows:
+                if not is_null(row[col]):
+                    value = row[col]
+                    break
+            merged_row[col] = value
+        #objs.append(_class(**merged_row))
+        objs.append(merged_row)
+    return objs, delete_stmt, None
```
bishax: Not reviewed this bundle of functions yet, as they're a little hard to follow (Lose track of types not being familiar with SQLalchemy etc.) and there's no tests to help.

bishax (commented):
I see the following 21 references to `health_data`:
* ./nesta/core/routines/projects/health_mosaic/nih/mesh_root_task.py:6:from nesta.core.routines.health_data.nih_data.mesh_join_task import MeshJoinTask
* ./nesta/core/routines/projects/health_mosaic/nih/tests/test_nih_process_task.py:4:# from nesta.core.routines.health_data.nih_data import nih_process_task
* ./nesta/core/routines/projects/health_mosaic/nih/nih_dedupe_task.py:21:from nesta.core.routines.health_data.nih_data.nih_abstracts_mesh_task import AbstractsMeshTask
* ./nesta/core/routines/projects/health_mosaic/nih/nih_dedupe_task.py:49:                                batchable=f3p("batchables/health_data/"
* ./nesta/core/routines/projects/health_mosaic/nih/mesh_join_task.py:17:from nesta.packages.health_data.process_mesh import retrieve_mesh_terms
* ./nesta/core/routines/projects/health_mosaic/nih/nih_collect_task.py:13:from nesta.packages.health_data.collect_nih import get_data_urls
* ./nesta/core/routines/projects/health_mosaic/nih/nih_root_task.py:17:from nesta.core.routines.health_data.nih_data.nih_dedupe_task import DedupeTask
* ./nesta/core/routines/projects/health_mosaic/nih/nih_root_task.py:48:                         batchable=f3p("batchables/health_data/"
* ./nesta/core/routines/projects/health_mosaic/nih/nih_abstracts_mesh_task.py:18:from nesta.core.routines.health_data.nih_data.nih_process_task import ProcessTask
* ./nesta/core/routines/projects/health_mosaic/nih/nih_abstracts_mesh_task.py:66:                          batchable=f3p("batchables/health_data/nih_process_data"),
* ./nesta/core/routines/projects/health_mosaic/nih/nih_process_task.py:16:from nesta.core.routines.health_data.nih_data.nih_collect_task import CollectTask
* ./nesta/core/routines/projects/health_mosaic/nih/nih_process_task.py:49:                          batchable=f3p("batchables/health_data/nih_collect_data"),
* ./nesta/core/routines/projects/health_mosaic/nih/nih_process_task.py:160:                          batchable=f3p("batchables/health_data/nih_process_data"),
* ./nesta/core/routines/projects/health_mosaic/companies/crunchbase_mesh_task.py:12:from nesta.packages.health_data.process_mesh import retrieve_mesh_terms, format_mesh_terms
* ./nesta/core/routines/datasets/nih/tests/test_nih_process_task.py:4:# from nesta.core.routines.health_data.nih_data import nih_process_task
* ./nesta/core/routines/datasets/nih/nih_root_task.py:22:from nesta.core.routines.health_data.nih_data.nih_dedupe_task import DedupeTask
* ./nesta/core/routines/datasets/nih/nih_root_task.py:53:                         batchable=f3p("batchables/health_data/"
* ./nesta/core/routines/datasets/nih/nih_process_task.py:21:from nesta.core.routines.health_data.nih_data.nih_collect_task import CollectTask
* ./nesta/core/routines/datasets/nih/nih_process_task.py:54:                          batchable=f3p("batchables/health_data/nih_collect_data"),
* ./nesta/core/routines/datasets/nih/nih_process_task.py:165:                          batchable=f3p("batchables/health_data/nih_process_data"),
* ./nesta/packages/nih/nih_synonyms.py:1:# from nesta.packages.health_data.nih_abstract_yielder import AbstractYielder


There are also a lot of other changes which I've tried to review where possible.

What's the reason for the apparent duplication of `nih_root_task`, `nih_dedupe_task` , `nih_process_tasks` (and accompanying tests)?

jaklinger:
> I see the following 21 references to health_data:

Thanks, fixed.

jaklinger:
> What's the reason for the apparent duplication of nih_root_task, nih_dedupe_task , nih_process_tasks (and accompanying tests)?

Apologies for the mess: these are placeholders which will be filled in in the subsequent PRs in this chain.

nesta/core/orms/mesh_orm.py
jaklinger: Yes: although useful for data integrity, they present a problem when you're dropping and developing the parent dataset. The mesh labelling system is anyway deprecated, so this table is just legacy as it stands.

nesta/core/orms/nih_orm.py
jaklinger: nice idea, thanks for upping my game

nesta/core/orms/nih_orm.py
jaklinger: autoincrement is one of the most annoying gotchas in sqlalchemy: it is always on by default, meaning that if you don't have a primary key present then it will assign you one. The problem with this is that when doing development you can end up with clashes, and also it doesn't "fail" if there is no PK present (which in this case there always should be).

I think just as a principle of good practice, we should always set this to False and then explicitly set to True as a way of communicating that this is an auto increment key. Seems crazy to me that this is the default behaviour!

nesta/core/orms/nih_orm.py
jaklinger: > performance implications

Takes up a little more disk (which is cheap), but since we have 2 million rows it will lead to a factor of 300k faster searching and joining. Since I'm writing some data-getters at the end of all this, figured that it would be useful!

jaklinger: > When to use VARCHAR vs TEXT?

Sorry, was going to comment on before changing from 'Draft' PR... Generally, I'm moving towards only using `VARCHAR` if the field needs to be indexed, otherwise use `TEXT` since it means we don't need to manage arbitrary `VARCHAR` sizes that are difficult to read. The only exception would be for short codes, where I think the `n` of `VARCHAR(n)` can give a little clarity (it would be worth indexing such an example anyway).

nesta/core/orms/nih_orm.py
jaklinger: Crazy, I know, but there is a journal title with 279 chars:

`The quarterly journal of nuclear medicine and molecular imaging : official publication of the Italian Association of Nuclear Medicine (AIMN) [and] the International Association of Radiopharmacology (IAR), [and] Section of the Society of Radiopharmaceutical Chemistry and Biology.`

I like to give a bit of leeway in case a bigger title somehow comes out in the future...

nesta/core/orms/nih_orm.py
jaklinger: so the field names and table names are automatically generated from the data collected from the NiH data dump, and so e.g. `pmid` is unfortunately their choice. Agreed that we need a data-dictionary per dataset. This is in scope for DAPS2, with the idea that we can introduce a new "`Column`" type which will also have a field called `description`, which will automatically end up in a separate `data_dictionary` table with fields like `table_name` `field_name` and `description`.

jaklinger: this will be in addition to the open ontology (e.g. schema.org), although it's to be discussed whether that's in scope for May 2021, or further down the line

nesta/core/orms/orm_utils.py
jaklinger: All of this `orm_utils` stuff is going to be factored into an intermediate PR before this Draft is "ready for review"

nesta/packages/nih/tests/test_nih_abstract_yielder.py
```diff
@@ -19,7 +19,7 @@ def setUp(self):
         session = Session()
         Base.metadata.drop_all(self.engine)
         Base.metadata.create_all(self.engine)
-        for i in range(0, self.n_abstracts):
+        for i in range(1, self.n_abstracts+1):
```
jaklinger, line 22: This is a consequence of doing `autoincrement=False`, since `0` is interpretted as meaning "increment me" by MySQL

jaklinger:
@bishax Thanks for the review. I've factored out the utils stuff, so there's nothing particularly controversial left in here now. I was going to factor out the `orms` as well, but seeing as you've already reviewed them I'll keep them in. I'll merge in the morning, if all tests should pass. 

There are two small PRs coming your way after this:

* the utils stuff (basically adding in a upsert method for dealing with nih data, and also a speed up for general insertions
* the nih raw data collection and preprocessing

PR #329, opened 2020-10-27
[326 precursor] Factored in utils from 326_rename

jaklinger:
Developed with to facilitate #326

All of changes here are required for `326_nih`, which concerns the (re)collection `nih` data, with various improvements in terms of data quality and pipeline running time, whilst factoring `nih` out of the `health-mosaic` project, for use in `eurito`.

A few utils for:
* Merging data by upsertion. There can be some ambiguity about what parts are being updated and how using raw SQL, and so instead I opt for 0) "manually" creating updated rows in python, based on the most recent non-null data 1) opening a transaction 2) dropping rows which are going to be updated 3) inserting the new rows
* This can be fairly slow, but it turns out using `insert(_class).values(chunk)` is 5x more performant than `session.bulk_save_objects(objs)` in the least impressive case and potentially 10-100x more performant depending on the number of rows being inserted. In the past this was causing a bottleneck for the NiH collection, but it looks like this is solved now.

Additionally:
* `bucket_keys` is introduced as a util, for retrieving all of the keys in a given bucket. It's a standard piece of boilerplate that I'm sick of.

Full set of tests add to cover the new features, plus you'll have to take my word that I've run the `arxiv`, `crunchbase`, `cordis`, `gtr` and `patstat` collections in `dev` mode and the new features haven't killed anything... (this is why we need end-to-end tests in DAPS2...). These uncovered a need for a small number of very minor updates before data is inserted into the database, for some of the oldest parts of the pipelines (`crunchbase`, `arxiv`, `gtr`)


nesta/core/luigihacks/misctools.py
```diff
@@ -125,3 +125,18 @@ def extract_task_info(luigi_task):
     task_name = type(luigi_task).__name__
     routine_id = f'{task_name}-{luigi_task.date}-{test}'
     return test, routine_id
+
+
+@lru_cache()
+def bucket_keys(bucket_name):
```
jaklinger, line 131: Note, this function is introduced to replace boilerplate in a bunch of luigi routines where batchables are marked as `done` if a given key exists (i.e. S3 is used as an intermediate checkpoint).

nesta/core/orms/orm_utils.py
```diff
@@ -341,6 +344,14 @@ def cast_as_sql_python_type(field, data):
     return _data
 
 
+def get_session(db_env, section, database, Base):
```
jaklinger, line 354: This used to be a part of `_filter_out_duplicates`, but is now used in both `_filter_out_duplicates` and the new `merge_duplicates`. Clearly factoring it out also allows for unit testing as well, as an added benefit

nesta/core/orms/orm_utils.py
```diff
@@ -458,23 +609,27 @@ def insert_data(db_env, section, database, Base,
...
-    try_until_allowed(Base.metadata.create_all, engine)
-    Session = try_until_allowed(sessionmaker, engine)
-    session = try_until_allowed(Session)
-    session.bulk_save_objects(objs)
+    # Open a transaction
+    session = get_session(db_env, section, database, Base)
+    if merge_non_null:
+        session.execute(existing_objs)  # Drop existing objs if merging
+    # Insert data in chunks
+    for chunk in split_batches(objs, insert_chunksize):
```
jaklinger, line 638: note: this is where `bulk_save_objects` (which takes an ORM instance as an argument) is replace by the SQL alchemy native `insert().values()`, which turns out to be significantly faster

nesta/core/batchables/gtr/collect_gtr/run.py
```diff
@@ -54,9 +55,10 @@ def run():
     objs = []
     for table_name, rows in data.items():
         _class = get_class_by_tablename(Base, f"gtr_{table_name}")
-        # Remove any fields that aren't in the ORM
-        cleaned_rows = [{k:v for k, v in row.items() if k in _class.__dict__}
-                        for row in rows]
+        # Remove any fields that aren't in the ORM and set NULL as default
```
jaklinger, line 58: The new `insert` method in `orm_utils` doesn't allow missing fields, so instead need to loop over fields in the table, rather than fields in the input data

nesta/core/batchables/gtr/collect_gtr/run.py
```diff
@@ -73,9 +75,8 @@ def run():
 if __name__ == "__main__":
     # Local testing
     if "BATCHPAR_outinfo" not in os.environ:
-        os.environ['BATCHPAR_TOPURL'] = "https://gtr.ukri.org/gtr/api/projects"
-        os.environ['BATCHPAR_PAGESIZE'] = "100"
-        os.environ['BATCHPAR_page'] = "647"
+        os.environ['BATCHPAR_PAGESIZE'] = "10"
```
jaklinger, line 78: local testing parameters

nesta/core/routines/datasets/crunchbase/crunchbase_org_collect_task.py
```diff
@@ -100,7 +100,7 @@ def run(self):
                                                                              low_memory=True)
         logging.info(f"Inserting {len(org_cats)} org categories "
                      f"({len(existing_org_cats)} already existed and {len(failed_org_cats)} failed)")
-        #org_cats = [OrganizationCategory(**org_cat) for org_cat in org_cats]
+        org_cats = [OrganizationCategory(**org_cat) for org_cat in org_cats]
```
jaklinger, line 103: legacy method of insertion from older pipelines

nesta/packages/arxiv/collect_arxiv.py
```diff
@@ -357,7 +357,8 @@ def add_new_articles(article_batch, session):
     logging.info(f"Inserting a batch of {len(article_batch)} new Articles")
     objs, existing_objs, failed_objs =  _filter_out_duplicates(session, Base, Article,
                                                                article_batch, low_memory=True)
-    session.add_all(objs)
+    articles = [Article(**obj) for obj in objs]
```
jaklinger, line 360: legacy method of insertion from older pipelines

nesta/packages/geo_utils/country_iso_code.py
```diff
@@ -25,8 +25,10 @@ def country_iso_code(country):
     country = str(country).title()
     for name_type in ['name', 'common_name', 'official_name']:
         query = {name_type: country}
-        try:
-            return pycountry.countries.get(**query)
+        try:            
+            result = pycountry.countries.get(**query)
+            if result is not None:
```
jaklinger, line 30: This is super daft, but an edge case came into the new data just *today* as I was running this end-to-end in `dev` mode that doesn't really warrant a PR. The upshot is that `pycountry` returns `None` for this value, which for us is equivalent to saying "not found".

PR #330, opened 2020-10-29
[326a] Deduplication with text vectors

jaklinger:
Refers to #326 

Generate doc vectors using the `Text2VecTask` task, and then run a FAISS indexer to generate a link table of exact- and near-duplicate abstracts and PHR fields.

Tasks:

- [x] Generate document vectors for PHR and abstract fields
- [x] Generate a FAISS index over the vectors for exact- and near-deduplication
- [x] Process the duplicates into a link table

This replaces the previous method of ingesting the data into one index on ES, and then running doc similarity before filling a second index on ES. This process was quite laborious, and also wastes an index and disk space. The bonus benefit of this method is that we get the doc vectors.

These haven't been chained together yet (this will happen in the final PR in this series), but the commands to run the two pipelines are:

```bash
luigi --module nih_vectors RootTask
```

and

```bash
luigi --module nih_dedupe_task RootTask
```

By "similarity score" ([defined in detail here](defined)), the following numbers of duplicates (without double-counting) of the NiH PHR field are:

```
+--------+----------+---------+---------+
| 0-0.55 | 0.55-0.6 | 0.6-0.8 | 0.8+    |
+--------+----------+---------+---------+
|  86176 |    52322 |  116581 | 1169913 |
+--------+----------+---------+---------+
```

And some examples are (note that I've literally done zero cherry picking here):

## Score == 0.8 (near duplicates)

```
*************************** 1. row ***************************
           fy: 2016
project_title: MARC Honors Undergraduate Research Training Program
project_start: 1978-06-01 00:00:00
  project_end: 2018-05-31 00:00:00
     org_name: City College Of New York
          phr: The goal of the proposed MARC U*STAR program is to further increase the number of qualified
 URM undergraduates who continue their education at the Ph.D. level in the biomedical sciences. This will
 enhance public health by increasing the pool of biomedical scientists working to study the underlying bases
 of biological phenomena and diseases.
*************************** 2. row ***************************
           fy: 2019
project_title: MARC Honors Undergraduate Research Training Program
project_start: 1978-06-01 00:00:00
  project_end: 2023-05-31 00:00:00
     org_name: City College Of New York
          phr: The goal of the proposed MARC U-STAR program is to further increase the number of qualified
 undergraduates from underrepresented groups who continue their education at the Ph.D. level in the 
biomedical and behavioral sciences. This will enhance public health by increasing the pool of biomedical
scientists working to study the underlying bases of biological phenomena and diseases.
```

## Score == 0.7 (Probably duplicates)
```
*************************** 1. row ***************************
           fy: 2016
project_title: Embryonic inheritance of sperm methylome after adult exposure to phthalates
project_start: 2016-08-01 00:00:00
  project_end: 2018-07-31 00:00:00
     org_name: University Of Massachusetts Amherst
          phr: This project investigates how adult exposures to phthalates affects DNA methylation of sperm
 and subsequently the embryo. The study has important implications for the effect of environmental 
endocrine disruptors in reproductive and offspring health.
*************************** 2. row ***************************
           fy: 2019
project_title: Male preconception phthalates and offspring embryo and sperm allele-specific methylome programming
project_start: 2017-09-01 00:00:00
  project_end: 2022-08-31 00:00:00
     org_name: University Of Massachusetts Amherst
          phr: This project investigates the effect of adult male exposures to phthalates on DNA methylation of 
sperm as well as embryonic tissue and germ cells of adult offspring. The study has important implications 
for the effect of environmental endocrine disruptors in reproductive and offspring health.
```

## Score == 0.6 (Contextually very similar)
```
*************************** 1. row ***************************
           fy: 2014
project_title: Novel Focal Adhesion Kinase autophosphorylation inhibitors against pancreatic cancer
project_start: 2014-09-24 00:00:00
  project_end: 2014-12-17 00:00:00
     org_name: Curefaktor Pharmaceuticals, Llc
          phr: Pancreatic cancer is a highly lethal disease with the worst prognosis among all cancer types and
 development of new treatments is critical for pancreatic cancer therapy. This application is focused on 
developing novel small molecule inhibitors targeting of Focal Adhesion Kinase (FAK) autophosphorylation to
 block pancreatic cancer cell growth and functions with the goal to identify best inhibitors for future clinical 
study. The preliminary data and this proposal provide a strong basis that novel inhibitors of FAK will have a 
significant impact on the therapeutic treatment of pancreatic cancer and on public health programs.
*************************** 2. row ***************************
           fy: 2019
project_title: Targeting LIF/LIFR in pancreatic cancer
project_start: 2019-04-02 00:00:00
  project_end: 2020-04-30 00:00:00
     org_name: Evestra, Inc.
          phr: Pancreatic cancer has one of the worst prognoses of all cancers due to intense stroma that 
impedes the perfusion of chemotherapeutics and impact therapy resistance to the cancer cells. The 
proposed research seeks to evaluate the therapeutic potential of small steroidal molecule EC359 in 
combination with gemcitabine that will simultaneously target pancreatic tumor stroma and cancer cells. The
 knowledge thus gained will provide a novel therapeutic regimen for the cure and management of pancreatic
 cancer patients.
```

## Score == 0.5 (Contextually fairly similar)

```
*************************** 1. row ***************************
           fy: 2014
project_title: Fathering Through Change: Online Parent Training for Divorced Fathers (FTC)
project_start: 2013-03-01 00:00:00
  project_end: 2016-07-31 00:00:00
     org_name: Iris Media, Inc.
          phr: Divorced fathers experience high rates of depression and anxiety due to parenting strains, and 
their children are at elevated risk for behavioral problems associated with divorce and co-parenting conflict. 
Evidence based parenting programs for fathers are lacking. The aim of this proposal is to develop a 
parenting program for divorced fathers to teach essential parenting skills associated with positive child 
outcomes and reduced conflict with former spouses. The societal benefits of the proposed program include 
higher rates of compliance with child support obligations and lower rates of litigation and re-litigation.
*************************** 2. row ***************************
           fy: 2015
project_title: Promoting Healthy Development among Children of Fathers with Antisocial Behavior
project_start: 2015-08-03 00:00:00
  project_end: 2017-06-30 00:00:00
     org_name: University Of Chicago
          phr: Children of fathers with a history of incarceration and antisocial behavior are at higher risk for 
conduct problems, substance abuse, mental health disorders, and poor academic achievement. The 
negative relation between fathers' imprisonment and antisocial behavior and children's outcomes calls for 
interventions to promote positive father engagement and parenting practices. The objective of this career
 development and research plan is to obtain high-quality training and to conduct advanced research to 
inform the development and feasibility testing of a family-focused intervention to reduce fathers' 
involvement in antisocial behavior and promote positive parenting to improve prospects for their children's
 well-being.
```


nesta/core/batchables/nlp/bert_vectorize/run.py
```diff
@@ -65,15 +65,29 @@ def run():
     # Convert text to vectors
     embeddings = model.encode(docs)
     # Write to output
-    out_data = [{"article_id": _id, "vector": embedding.tolist()}
```
jaklinger, line 68: this was a bug from the `arxiv` vectorisation pipeline (which was the only case that `bert_vectorize` was used in) where `id_field="article_id"`, so just replaced with the generic variable `id_field` as required

nesta/core/batchables/nlp/bert_vectorize/run.py
```diff
@@ -65,15 +65,29 @@ def run():
...
+    out_data = [{id_field: _id, "vector": ["%.5f" % v for v in embedding.tolist()]}
                 for _id, embedding in zip(ids, embeddings)]
     Base = get_base_from_orm_name(out_module)
     out_class = get_class_by_tablename(out_module, out_tablename)
     insert_data("BATCHPAR_config", "mysqldb", db_name, Base,
-                out_class, out_data, low_memory=True)
+                out_class, out_data, low_memory=True,
+                insert_chunksize=10)
```
jaklinger, line 74: `insert_chunksize` is a new feature, so "upgrading" the insertion which minimises bottlenecks with MySQL due to very large chunks (I've found that 10, 100, 1000 all have the same run time, so settled on the smallest)

nesta/core/batchables/nlp/bert_vectorize/run.py
```diff
@@ -65,15 +65,29 @@ def run():
     # Convert text to vectors
     embeddings = model.encode(docs)
     # Write to output
-    out_data = [{"article_id": _id, "vector": embedding.tolist()}
+    out_data = [{id_field: _id, "vector": ["%.5f" % v for v in embedding.tolist()]}
```
jaklinger, line 68: Truncating the vectors to 5 decimal places, rather than 20 to save on bytes just for unnecessary (and also illegible) accuracy. This also makes the ingestion a little quicker.

nesta/core/orms/orm_utils.py
```diff
@@ -630,17 +630,23 @@ def insert_data(db_env, section, database, Base,
                                data=data,
                                low_memory=low_memory)
     objs, existing_objs, failed_objs = response
-    # Open a transaction
-    session = get_session(db_env, section, database, Base)
-    if merge_non_null:
-        session.execute(existing_objs)  # Drop existing objs if merging
```
jaklinger, line 636: removed in favour of `db_session` context manager. The reason for this (other than it being cleaner), is explained below

nesta/core/orms/orm_utils.py
```diff
@@ -630,17 +630,23 @@ def insert_data(db_env, section, database, Base,
...
+            session.execute(existing_objs)
+
     # Insert data in chunks
     for chunk in split_batches(objs, insert_chunksize):
-        stmt = insert(_class).values(chunk)
-        session.execute(stmt)
-    # Commit and close
-    session.commit()
-    session.close()
+        with db_session(engine) as session:
```
jaklinger, line 645: if the total length of time of the transaction is too long (as is the case when inserting several million 700-dim vectors) then a single session will die eventually. Using smaller transactions is the only way forward, and doesn't affect other pipelines, and is anyway better practice.

nesta/core/scripts/nesta_docker_build
```diff
@@ -73,7 +73,8 @@ cd ${TOP_DIR}
 DKR_IMG_NAME=${AWS_DKR_PATH}/${DKR_PKG_NAME}:latest
 docker tag ${DKR_PKG_NAME}:latest ${DKR_IMG_NAME}
 docker push ${DKR_IMG_NAME}
-docker system prune -a -f
```
jaklinger, line 76: this was a bit heavy-handed, causing all images (including those in production) to be deleted...

nesta/core/luigihacks/sql2batchtask.py
```diff
@@ -43,6 +44,11 @@ class Sql2BatchTask(autobatch.AutoBatchTask):
     process_batch_size = luigi.IntParameter(default=1000)
     id_field = SqlAlchemyParameter()
     filter = SqlAlchemyParameter(default=None)
+    # Note: ParameterVisibility set for "filter_ids" because 
+    # the list of filter IDs can be arbitarily long, and so it clogs up
+    # the i/o stream and the email error message
+    filter_ids = luigi.ListParameter(default=[],
+                                     visibility=ParameterVisibility.PRIVATE)
```
bishax, line 51: To check: setting `PRIVATE` doesn't change the behaviour of the task?

nesta/core/orms/nih_orm.py
```diff
@@ -120,3 +119,37 @@ class ClinicalStudies(Base):
     core_project_number = Column(VARCHAR(50), index=True)
     study = Column(TEXT)
     study_status = Column(VARCHAR(30), index=True)
+
+
+class PhrVector(Base):
+    """Document vectors for NiH PHR statements."""
```
bishax: Docs: give acronym for PHR

nesta/core/orms/nih_orm.py
```diff
@@ -120,3 +119,37 @@ class ClinicalStudies(Base):
...
+    __tablename__ = 'nih_phr_vectors'
+    application_id = Column(INTEGER, ForeignKey(Projects.application_id),
+                            autoincrement=False, primary_key=True)
+    vector = Column(JSON)
+
+
+class AbstractVector(Base):
+    """Document vectors for NiH abstracts."""
+    __tablename__ = 'nih_abstract_vectors'
+    application_id = Column(INTEGER, ForeignKey(Abstracts.application_id),
```
bishax, line 135: Are the two `application_id` distinct entities? I'm guessing it's naming carried over from NiH?

nesta/core/orms/nih_orm.py
```diff
@@ -120,3 +119,37 @@ class ClinicalStudies(Base):
...
+
+
+class TextDuplicate(Base):
+    """Link table to describe for NiH text-field duplicates,
+    which probably imply that projects are related, either
+    formally (if weight > 0.8 they are normally 
+    almost exact duplicates of each other) or contextually 
+    (if weight > 0.5 it is normally in the same general
+    subject area).
+    """
```
bishax: Docs: Definitions here are less detailed than the PR summary. Is the cut-off for inclusion in this table 0.5 or something else?

nesta/core/routines/datasets/nih/nih_dedupe_task.py
```diff
@@ -1,131 +1,75 @@
...
 
-from nesta.core.routines.nih.nih_data.nih_abstracts_mesh_task import AbstractsMeshTask
+from nesta.packages.vectors.similarity import generate_duplicate_links
+import faiss
 
-class DedupeTask(autobatch.AutoBatchTask):
-    '''
+
+class VectorDedupeTask(luigi.Task):
+    '''Generic task for generating a link table duplicate text fields
```
bishax, line 27: If this is a generic task should it go in a generic module?

bishax: Looking below, I guess it's not really generic as it's mostly a NiH specific wrapper around the generic `generate_duplicate_links`

nesta/core/routines/datasets/nih/nih_dedupe_task.py
```diff
@@ -1,131 +1,75 @@
...
+    
+    def requires(self):
+        for source, orm in (("phr", PhrVector),
+                            ("abstract", AbstractVector)):
+            yield VectorDedupeTask(vector_orm=orm,
+                                   source=source,
+                                   date=self.date,
+                                   test=not self.production)
+            if not self.production:
                 break
```
bishax, line 75: Could this mask errors in deduping abstracts when testing?

nesta/core/scripts/docker_recipes/py36_amzn1_image/launch.sh
```diff
@@ -26,6 +26,7 @@ ls
 # Remove the following from reqs since they're huge
 sed -i '/luigi/d' requirements.txt
```
bishax, line 27: As the number of images and packages to exclude from given images grows, this could get more confusing.
Might be worth specifying a generic blacklist of packages (any `sed`'ed packages) and a list of images which allow given packages from the blacklist.
Then each image can run the same script that takes the image name & blacklist then returns packages to `sed` out.

Possibly not worth the effort?

nesta/core/scripts/docker_recipes/py37_amzn2_pytorch/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+RUN yum install gcc -y
+RUN yum clean all -y
+
+# Install base conda env
+RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh && /bin/bash ~/miniconda.sh -b -p /opt/conda && rm ~/miniconda.sh && ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh && echo ". /opt/conda/etc/profile.d/conda.sh" >> ~/.bashrc && echo "conda activate base" >> ~/.bashrc && /bin/bash ~/.bashrc
+ENV PATH /opt/conda/bin:$PATH
+RUN conda config --add channels conda-forge
+RUN conda create -y -n py37 \
+    python=3.7 \
+    sentence-transformers \
```
bishax, line 28: Any reason for installing with conda over pip?

nesta/packages/vectors/read.py
```diff
@@ -0,0 +1,109 @@
...
+    data, ids = prefill_inputs(orm, database)  # Empty numpy arrays
+    while "reading data":
+        try:
+            # Start or continue reading
+            read_data(data=data, ids=ids, orm=orm, 
+                      id_field=id_field, database=database,
+                      chunksize=chunksize, 
+                      max_chunks=max_chunks)
+        # The following has only been found to happen if your
+        # connection drops slightly, which corrupts the JSON
```
bishax, line 112: How does the corruption happen?

bishax (commented):
Looking good mostly.
- `nih_vectors` is `:(` because querying for `AbstractVector.application_id` in `nih_vectors.py:41` returns `[]` which can't be unpacked by `done_ids, =`
- I had to do the following to get `nih_dedupe_task` to run: `conda install faiss-cpu -c pytorch`

nesta/core/luigihacks/sql2batchtask.py
jaklinger: Literally no difference in behaviour, I guess it was intended for sensitive info but in this case if you're passing in a very large list it's just annoying for it to be visible in the output!

nesta/core/orms/nih_orm.py
jaklinger: it's pretty daft: they *are* the same and *in principle* `nih_abstracts.application_id` should be the subset of `nih_projects.application_id`, however there are around 128 (less than 0.011%) of `nih_abstracts.application_id` which don't have a matching `nih_projects.application_id` - meaning that an FK or relationship can't be set...

nesta/core/routines/datasets/nih/nih_dedupe_task.py
jaklinger: Exactly, there are elements of it which are generic but this dataset is already an exceptional case because there are two text fields in two separate tables which could indicate being a duplicate. Perhaps with more use-cases it will be easier to factor out commonalities, but in the meanwhile the `generate_duplicate_links` is pretty generic

nesta/core/routines/datasets/nih/nih_dedupe_task.py
jaklinger: the main reason for this is that it's quite a slow turnaround time: but you would only expect errors to crop up if there are e.g. weird formattings etc in abstracts compared with PHR, but having already processed *everything* in `production` mode in anticipation of this PR I can already say that there really aren't any edge cases of that type. For me, the purpose of the `test/production` modes is something between unit and end-to-end testing, to at least check that the basic principle of the pipeline is working - which I think is covered in this case as `PhrVector` and `AbstractVector` are so similar.

jaklinger:
> nih_vectors is :( because querying for AbstractVector.application_id in nih_vectors.py:41 returns [] which can't be unpacked by done_ids, =

Ah thanks for catching this. "Fixed" it with an `except ValueError` for this case.

jaklinger:
> faiss-cpu -c pytorch

Thanks, updated the reqs

nesta/core/scripts/docker_recipes/py37_amzn2_pytorch/Dockerfile
jaklinger: not really, I was always under the impression that the dependency management was done better but I never really tested the difference. It works as is: but do you see a reason to change it? (e.g. bloatiness?)

nesta/packages/vectors/read.py
jaklinger: A while back I gave a prototype of the above code to Juan and we both had issues with JSON corruption, which I found went away by using ethernet. Debugging with an `except`, I found that a single character is occasionally altered in the middle of the JSON leading to bad JSON parsing. I can only assume that somehow it's related to how the data is unpacked over WiFi... seems like a potential security flaw to me, but I'm not sure I want to dig that deep!

nesta/core/scripts/docker_recipes/py37_amzn2_pytorch/Dockerfile
bishax: Conda is slow and always seems to break but has always been something to put up with but now I struggle to see the point in it anymore...
Python wheels remove the need for conda in a lot of cases, and when there aren't wheels then we can get consistent system dependencies in Docker anyway.

Furthermore, `pip install`'ing on top of a conda environment undoes a lot of the conda dependency management anyway because `pip` will fetch the versions it wants for the packages it has been told to store.

bishax: I say this as someone who always advocated for Conda because Docker has a higher barrier to entry, but the amount of time I lose in the average month due to conda randomly breaking / slow dependency resolution / slow rebuilding of environments etc. can be quite significant.

And lets not even go into

nesta/core/routines/datasets/nih/nih_drop_abstracts.py
```diff
@@ -0,0 +1,63 @@
...
+        database = 'dev' if self.test else 'production'
+        engine = get_mysql_engine("MYSQLDB", "mysqldb", database)
+        with db_session(engine) as sess:
+            # Query missing application ids
+            id_ =  Abstracts.application_id
+            join_ = id_ == Projects.application_id
+            filter_ = Projects.application_id == None
+            q = sess.query(id_,).outerjoin(Projects, join_).filter(filter_)
+            # Drop the missing ids
+            dud_ids, = zip(*q.all())
```
bishax: Unpacking error similar to `nih_vectors.py:41`

nesta/core/scripts/docker_recipes/py37_amzn2_pytorch/Dockerfile
jaklinger: ok, fair criticism. I've put this into `daps_utils` as an [issue](https://github.com/nestauk/daps_utils/issues/32) so that the default env doesn't have `conda`. Migrating from `nesta/nestauk` to `mono_daps` (or whatever we call it) will start in the next few months and so we can ditch at that point. Sound ok?

The docker setup in `daps_utils` is slimmer as well, as it uses a base image so that none of the duplication/boilerplate stuff is needed

nesta/core/scripts/docker_recipes/py36_amzn1_image/launch.sh
jaklinger: I would say not worth the effort *because* in new daps, the `requirements.txt` is always relative a specified `flow`, and so bloaty reqs should go in their local repo rather than clogging it up for everyone else. This kind of stuff was one of the motivations for factoring out mono-repo reqs into a per-project basis

bishax (commented):
[Apologies if premature re-review]

### 1

> > nih_vectors is :( because querying for AbstractVector.application_id in nih_vectors.py:41 returns [] which can't be unpacked by done_ids, =

> Ah thanks for catching this. "Fixed" it with an except ValueError for this case.

This fix doesn't look like it's been committed (noted similar error for drop abstracts task)

### 2

The fix to requirements is slightly incorrect `pytorch` should be `torch` because the conda and pip packages are named differently. Testing this in a fresh environment led me to discover that I don't get `mysql_connector_python` which means tasks fail, I'm not sure what the deal with `mysql_connector_repackaged` is (guessing I've missed something)?
With the `torch` correction and `pip install mysql_connector_python` then `nih_dedupe_task` is `:)`

nesta/core/scripts/docker_recipes/py37_amzn2_pytorch/Dockerfile
bishax: I think https://python-poetry.org/ is possibly worth a look into

jaklinger:
@bishax Yes sorry I was nearly about to commit those changes - sorry got caught up in another PR...

Just pushed the changes (you can [see all changes from the last couple of days here](https://github.com/nestauk/nesta/pull/330/files/7c19f0c1b16e6211c4938c1ea38c47d951872cf8..ae6ad9316f1d5a1d3e8c4f359f29549596e70607))

bishax (approved):
Apart from the outstanding comment about `mysql_connector_python` (which doesn't strictly relate to this PR) then good to merge!

PR #337, opened 2020-11-19
[326] NiH: impute base id

jaklinger:
Towards addressing #326 

What NiH don't tell you is that the `core_project_num` field has a "base" project number, which is effectively the actual core project number. This is essential for aggregating projects, otherwise it will appear that many duplicates exist in the data.

Can test the functionality by doing:

`luigi --module nih_impute_base_id RootTask`

from `nesta/core/routines/datasets/nih`

### DIFF
--- a/nesta/core/routines/datasets/nih/nih_impute_base_id.py
+++ b/nesta/core/routines/datasets/nih/nih_impute_base_id.py
@@ -1,0 +1,57 @@
+'''
+Impute Base ID
+==============
+
+What NiH don't tell you is that the `core_project_num` field
+has a base project number, which is effectively the actual
+core project number. This is essential for aggregating projects,
+otherwise it will appear that many duplicates exist in the data.
+
+This task imputes these values using a simple regex of the form:
+
+    {BASE_ID}-{an integer}-{an integer}-{an integer}
+
+Any `core_project_num` failing this regex are ignored.
+'''
+
+import logging
+import luigi
+from datetime import datetime as dt
+from multiprocessing.dummy import Pool as ThreadPool
+
+from nesta.core.luigihacks.mysqldb import make_mysql_target
+from nesta.packages.nih.impute_base_id import impute_base_id_thread
+
+
+class ImputeBaseIDTask(luigi.Task):
+    '''Impute the base ID using a regex of the form
+    {BASE_ID}-{an integer}-{an integer}-{an integer}
+
+    Args:
+        date (datetime): Date stamp.
+        test (bool): Running in test mode?
+    '''
+    date = luigi.DateParameter()
+    test = luigi.BoolParameter()
+
+    def output(self):
+        return make_mysql_target(self)
+
+    def run(self):
+        database = 'dev' if self.test else 'production'
+        id_ranges = retrieve_id_ranges(database)
+        # Threading required because it takes 2-3 days to impute
+        # all values on a single thread, or a few hours on 15 threads
+        pool = ThreadPool(15)
+        results = pool.map(impute_base_id_thread, id_ranges)
+        pool.close()
+        pool.join()
+        self.output().touch()
+
+
+class RootTask(luigi.WrapperTask):
+    date = luigi.DateParameter(default=dt.now())
+    production = luigi.BoolParameter(default=False)
+
+    def requires(self):
+        yield ImputeBaseIDTask(date=self.date, test=not self.production)
--- a/nesta/core/routines/datasets/nih/nih_impute_base_id.py
+++ b/nesta/core/routines/datasets/nih/nih_impute_base_id.py
@@ -14,9 +14,11 @@ This task imputes these values using a simple regex of the form:
 Any `core_project_num` failing this regex are ignored.
 '''
 
+
 import logging
 import luigi
 from datetime import datetime as dt
+from itertools import repeat
 from multiprocessing.dummy import Pool as ThreadPool
 
 from nesta.core.luigihacks.mysqldb import make_mysql_target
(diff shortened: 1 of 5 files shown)


nesta/packages/nih/impute_base_id.py
```diff
@@ -0,0 +1,95 @@
...
+    over which "impute_base_id" can be mapped"""
+    engine = get_mysql_engine("MYSQLDB", "mysqldb", database)
+    # First get all offset values
+    with db_session(engine) as session:
+        q = session.query(Projects.application_id)
+        q = q.order_by(Projects.application_id)
+        try:
+            ids, = zip(*q.all())
+        except ValueError:  # Forgiveness, if there are no IDs in the DB
+            ids = []
```
bishax: If  `ids = []` then the next line will fail

nesta/core/routines/datasets/nih/nih_impute_base_id.py
```diff
@@ -0,0 +1,58 @@
...
+    def output(self):
+        return make_mysql_target(self)
+
+    def run(self):
+        database = 'dev' if self.test else 'production'
+        id_ranges = retrieve_id_ranges(database)
+        # Threading required because it takes 2-3 days to impute
+        # all values on a single thread, or a few hours on 15 threads
+        pool = ThreadPool(15)
+        results = pool.map(impute_base_id_thread, id_ranges)
```
bishax: `pool.starmap(impute_base_id_thread, id_ranges, database)` stays true to naming

nesta/packages/nih/impute_base_id.py
```diff
@@ -0,0 +1,95 @@
...
+    final_id = ids[-1]
+    ids = list(ids[0::chunksize])  # Every {chunksize}th id
+    # Pop the final ID back in, if it has been truncated
+    if ids[-1] != final_id:
+        ids.append(final_id)
+    # Zip together consecutive pairs of arguments, i.e.
+    # n-1 values of (from_id, to_id, database)
+    # where from_id[n] == to_id[n-1]
+    n = len(ids) - 1
+    id_ranges = list(zip(ids, ids[1:], [database]*n))
```
bishax: database doesn't need to be included here, see `starmap` suggestion in luigi task file

bishax (commented):
One bad codepath and a small suggestion

nesta/core/routines/datasets/nih/nih_impute_base_id.py
jaklinger: I'm not sure if this is how `starmap` works - from what I can see, by swapping from `map` to `starmap` I would keep everything the same (including the ugly repeated `database` arg) except I would change:

```python
def impute_base_id_thread(*args)
```

to 

```python
def impute_base_id_thread(from_id, to_id, database)
```

jaklinger: since the pattern for `starmap` is:

```python
pool.starmap(func, iterable[, chunksize])
```

which is the same as `map`. Or have I missed something?

nesta/packages/nih/impute_base_id.py
jaklinger: Thanks, I'll just `return []`  instead!

nesta/core/routines/datasets/nih/nih_impute_base_id.py
bishax: You haven't missed something, my brain short-circuited.
I've often used `starmap` in combination with `itertools.product`, and ended up merging the two into one step.

What that would have suggested was:
```
from itertools import product
pool.starmap(impute_based_id_thread, product(id_ranges, database))
```

But the following is simpler (given `database` only ever takes the same value)
```
from itertools import repeat
pool.starmap(impute_based_id_thread, zip(id_ranges, repeat(database)))
```

Then you'd have `def impute_base_id_thread(id_range, database)`.

jaklinger: actually in the end this doesn't work either as you need to unpack and then chain the `id_ranges` which have already been generated. One solution is:

```python
from itertools import chain

_id_ranges = map(lambda x: chain(x, [database]), id_ranges)
pool.starmap(impute_base_id_thread, _id_ranges)
```

which works nicely and is also purely iterable. I looked into using `partial` rather than the lambda, but it won't work since with `partial` unnamed args are parsed first and the `database` arg in my `chain` has to go last. No biggy, the lambda doesn't look too bad!

bishax: Sorry to continue discussing this but I like little exercises like this :sweat_smile: 

To clarify I was still suggesting unpacking the range within the function (as below).
The suggested change to starmap was mostly about a clearer mapping between function/variable naming and what the code did: `retrieve_id_ranges` returns a range of id's (without added information about `database`); and the variable `id_ranges` were actually id ranges rather than id ranges and a `database` label.

```
from itertools import repeat
import multiprocessing.dummy as mpd

def foo(id_range, database):
    start, end = id_range
    print(f"{start}-{end}:{database}")

# Dummy data
database = "laterbase"
id_ranges = [(0, 2), (2, 3), (3, 3_00), (4, 5)]

pool = mpd.Pool()
pool.starmap(foo, zip(id_ranges, repeat(database)))

# OUTPUT
0-2:laterbase
2-3:laterbase
3-300:laterbase
4-5:laterbase
```

PS. not saying it should be changed or anything, just wanted to make my intentions clearer

jaklinger: Gotcha, thanks!

Issue #347, opened 2020-12-02
[DAPS] gtrans fails


jaklinger:
See #https://github.com/ssut/py-googletrans/issues/234#issuecomment-737229318 for potential fix

PR #350, opened 2020-12-22
[347] Turn off translate

jaklinger:
Refers to #347

Disables translation, until fix is implemented


jaklinger:
confirmed that arXlive runs again. Nothing controversial so merging until fix is implemented.

